Issue #320, opened 2014-12-24
CFD-205: Field widgets : Select Topic


nader77:
2-level topic selection widget.
Selecting the subtopic is only possible when the parent levels are also selected.
Entity reference (node) field

zero2one:
Same problem as #325 : Unselecting the parent item does not unselect the child terms.

zero2one:
QC OK.

PR #322, opened 2014-12-24
CFD-205: Field widgets : Select Topic

nader77:
#320


nader77:

Issue #321, opened 2014-12-24
CFD-208: Field widgets : Date selection


nader77:
1 level taxonomy selection.
Entity reference (taxonomy) field

zero2one:
Merged

PR #324, opened 2014-12-24
CFD-208: Field widgets : Date selection.

nader77:
#321


nader77:

Issue #325, opened 2014-12-25
CFD-206: Field widgets : Select Regions & Countries


nader77:
Up to 3 level taxonomy.
- Region
   - Subregion
     - Country
- Region
  - Country
- Region
Selecting a term is only possible if the parent terms are also selected.

Entity reference (taxonomy) field

zero2one:
Possible to select countries without the parent region.
  
## Test workflow:
![c4m-taxonomy-region-country](https://cloud.githubusercontent.com/assets/133124/5642875/2160a308-9651-11e4-820d-15b0a4fcc4e7.gif)

## Test result:
![c4m-taxonomy-region-country](https://cloud.githubusercontent.com/assets/133124/5642880/33c9be1c-9651-11e4-8322-6e0de8825007.png)

## Test Scenario:
  - Create new discussion.
  - Select "7. EU Member states"
  - Select "Belgium", "Bulgaria", "Croatie"
  - Close the checkbox popup.
  - Remove "7. EU Member states"
  - Save 

#### Unexpected result:
- The EU memberstate is not listed (OK)
- "Belgium", "Bulgaria", "Croatie" are listed (NOK)
 
#### Desired result:
- removing the parent region should also remove the child countries.

zero2one:
QC OK, merged.

PR #326, opened 2014-12-25
CFD-206: Field widgets : Select Regions & Countries

nader77:
#325


nader77:

Issue #327, opened 2014-12-25
Separate the AngularJs App controllers.


nader77:
* Avoid unnecessary restful calls.
* Minimize and re-order the AngularJs code.
* Call the angularJs app only when necessary.

PR #328, opened 2014-12-25
Separate the AngularJs App controllers.

nader77:
#327 
* Created a new controller called "Activity". (This will eliminate any unnecessary calls to activity-stream)
* Created a broadcast that communicates between the 2 controllers.
* Fixed the scroll when creating a new activity.
* Optimized the code in the main controller, Because we need to call it in many places, this will make it faster to load.

Adding new activity with the new fixes:
activity


nader77:
Update: optimized the refresh functions to avoid any conflicts between the normal refresh of the activity stream and the data pull of creating a new Activity.

nader77:
@amitaibu This PR is ready for code review.

amitaibu:
Please open PR against ``develop-widgets``, and close this.

nader77:
This is merged into ``develop-widgets``

Issue #332, opened 2014-12-30
CFD-244 : Add Tags autocomplete widget to full form.


nader77:

zero2one:
Merged.

PR #333, opened 2014-12-30
Add Tags autocomplete widget to full form.

nader77:
#332


nader77:

PR #338, opened 2015-01-01
Fix "ESC" close popups for taxonomy.

nader77:
#337


nader77:
Update: Fixed the popup closing on ESC key press in both full form and quick post.
The only way was a duplicated code in each controller, The other option (Calling a service) required a big amount of code in each controller which means bigger duplicated code.

Workflow:
popups

nader77:
@amitaibu ^^ 

Any better suggestions ?

Issue #354, opened 2015-01-05
Drop file.


nader77:
* Can't see the "Drop file here to upload", because its in white. 
![selection_052](https://cloud.githubusercontent.com/assets/7369740/5610931/ed4ee78a-94c7-11e4-8b54-3b9cb36d6232.png)

* When dropping a file it's not working as expected, Redirects to an empty page:
![selection_053](https://cloud.githubusercontent.com/assets/7369740/5610937/138dfa44-94c8-11e4-904a-2eaf3b7907fa.png)

zero2one:
QC OK, merged.

PR #355, opened 2015-01-05
Drop file.

HelenaEksler:
#354


HelenaEksler:
@nader77 For now we are creating discussion only inside some group - so we should have group context.

selection_158

Text in the widget is not white any more.

selection_159

HelenaEksler:
@nader77 @amitaibu PR is ready.

HelenaEksler:
Closed.

Issue #360, opened 2015-01-06
CFD-210: Full form : Discussion: Re-order form elements.


nader77:

zero2one:
Desired field order:
  - Title
  - Type of discussion
  - Description
  - Documents
  - Topics
  - Group categories
  - Date
  - Language
  - Regions & Countries
  - Tags

zero2one:
QC OK, merged.

PR #361, opened 2015-01-06
CFD-210: Full form : Discussion: Re-order form elements.

nader77:
#360


nader77:
@amitaibu Essentially the problem in the order is that the tags are under the categories, because they are made in the same widget, I separated them into 2 markups:
selection_055

This fixes the order of the "tags" for all the content types.

HelenaEksler:
Closed.

PR #368, opened 2015-01-06
CFD-224 - Don't show the global menu when in group context

soniCaH:


soniCaH:
groucontext
globalcontext

PR #371, opened 2015-01-07
Uncheck children terms when parent term is unchecked.

HelenaEksler:
#374


HelenaEksler:
taxfilter2

HelenaEksler:
@amitaibu PR is ready.

PR #373, opened 2015-01-07
CFD-210: Full form : Discussion: Re-order form elements.

HelenaEksler:
#360


HelenaEksler:
@amitaibu Essentially the problem in the order is that the tags are under the categories, because they are made in the same widget, I separated them into 2 markups:
selection_055

This fixes the order of the "tags" for all the content types.

HelenaEksler:
@zero2one in the Event form - tags field should be inside fieldset or outside?
selection_163

HelenaEksler:
Event page: right fields order.
create event capacity4more

Issue #376, opened 2015-01-07
CFD-215 : Group Library : Add field to document type to determine if document should be listed in the library.


zero2one:
Users can upload documents while creating group content (discussion, event, wiki page, …).

A form will open that gives them the option to select if the uploaded file should be listed in the document library.

Add a field to the Document type (checkbox) to set if the document should be listed in the document library.
The default value should be *YES* (listed).

This checkbox should be used in the file widget to store if a user wants the document to be listed in the library.

zero2one:
QC OK, merged.

PR #377, opened 2015-01-07
WIP: CFD-215 : Group Library : Add field to document type to determine if document should be listed in the library.

HelenaEksler:
#376


ckorakas:
:+1:

Issue #385, opened 2015-01-08
CFD-250 : c4m-app should only be loaded on pages where it is needed


amitaibu:

HelenaEksler:
@amitaibu  Could you add some description of what should be done here?

zero2one:
Right now, the c4m-app (angular) is loaded on every page + the body attributes are altered.
Thats not needed.
Can we:
Create a c4m_app_init_js() function that:
Sets the proper body tag attributes
Runs the needed drupal_add_js & drupal_add_css functions to load the necessary dependencies.
Call that function everywhere where the functionality (quick post, form field widgets) requires it.

zero2one:
QC OK, Merged.

PR #391, opened 2015-01-08
CFD-250 : App should only be loaded on pages where it is needed 

HelenaEksler:
#385


HelenaEksler:
@amitaibu PR is ready for the code review.

capacity4more/modules/c4m/restful/c4m_restful_quick_post/c4m_restful_quick_post.module
```diff
@@ -163,10 +163,18 @@ function c4m_restful_quick_post_c4m_location_field_entity_property_info(&$info,
  * Attach Angular application to the page.
  */
 function c4m_restful_quick_post_preprocess_html(&$vars) {
+  drupal_static_reset('menu_get_item');
```
amitaibu: looks wrong.

HelenaEksler:
@amitaibu Can we assign this to Peter?

zero2one:
Page is still loaded on the site homepage (/).

It should only be loaded on the group dashboard IF the quick post form is shown.

Only group members should see the quick post on the group dashboard.

zero2one:
Do we really need the c4m-app on the detail page?

capacity4more/modules/c4m/restful/c4m_restful_quick_post/c4m_restful_quick_post.module
```diff
@@ -165,8 +165,15 @@ function c4m_restful_quick_post_c4m_location_field_entity_property_info(&$info,
 function c4m_restful_quick_post_preprocess_html(&$vars) {
   $item = menu_get_item();
 
-  if (!menu_get_object() && isset($item['map']) && $item['map'][0] != 'node') {
-    // Not a node page and not a node/add(edit) page.
+  $allowed_paths = array(
+    'node/add',
+    'node/%/edit',
+    'node',
```
zero2one: We don't need c4m-app on the node page.

HelenaEksler: Pages with menu item path 'node' are a groups' dashboards. So we need app on this page. For example, on some discussion page application is not loading.
screenshot from 2015-01-20 15 55 17

HelenaEksler:
@zero2one 
And also we need application even if user doesn't see quick post form in the dashboard page, because the activity stream is also loading within c4m-app.

HelenaEksler:
@zero2one 
>Only group members should see the quick post on the group dashboard.

And what about administrators?

zero2one:
@HelenaEksler 

* Don't bother the access yet.
* You are right: c4m-app is needed for the activity stream.

Can you build in an extra check that we load the c4m-app only on:
* OR path = node/add/%
* OR path = node/%/edit
* OR (path = <front> AND OG context exists)

Don't use node to check for front page as this may change…

Se we don(t load c4m-app on the site front page?

capacity4more/modules/c4m/restful/c4m_restful_quick_post/c4m_restful_quick_post.module
```diff
@@ -165,8 +169,19 @@ function c4m_restful_quick_post_c4m_location_field_entity_property_info(&$info,
...
+  );
+
+  if (!in_array($item['path'], $allowed_paths)) {
+    // Not a group dashboard page and not a node/add(edit) page.
+    return;
+  }
+  if ($item['path'] == 'node' && !og_context()) {
```
zero2one: Is there a good reason not to use the [drupal_is_front_page()](https://api.drupal.org/api/drupal/includes%21path.inc/function/drupal_is_front_page/7) function?

```
if (drupal_is_front_page() && !og_context()) {
  ...
}
```

HelenaEksler:
@zero2one  I agree. So using ``drupal_is_front_page()``.

capacity4more/modules/c4m/restful/c4m_restful_quick_post/c4m_restful_quick_post.module
```diff
@@ -164,9 +168,19 @@ function c4m_restful_quick_post_c4m_location_field_entity_property_info(&$info,
...
+    'node',
+  );
 
-  if (!menu_get_object() && isset($item['map']) && $item['map'][0] != 'node') {
-    // Not a node page and not a node/add(edit) page.
+  if (!in_array($item['path'], $allowed_paths)) {
+    // Not a group dashboard page and not a node/add(edit) page.
+    return;
+  }
+  if (drupal_is_front_page() && !og_context()) {
```
zero2one: You need to refactor the code more:
* If we ever decide to disable the /node page (and we should).
* If we ever decide that we use another default home page (e.g. /home).

Then the c4m-app will no longer load on the group dashboard because the path (/home) no longer in the allowed_paths list.

Maybe you need to change the allowed paths list like this:
```
$allowed_paths = array(
  'node/add',
  'node/%/edit',
  variable_get('site_frontpage', 'node')
);
```

Should the add path not be like:
```
  'node/add/%'
```

We don't need the app on `node/add` itself as this is an overview of all possible content type…

HelenaEksler:
@amitaibu Could you review this, please?

amitaibu:
needs a re-roll

PR #393, opened 2015-01-09
WIP: cfd 216 group library document teaser display

jummonk:


jummonk:
library list
library table

PR #397, opened 2015-01-09
cfd 179 group events overview page upcoming past

jummonk:


jummonk:
my4je1sglu

zero2one:
The overview page should not show any pager (upcomming / past) when there are more then 5 items in one of the 2 lists.

c4m-events-overview-no-pager

PR #399, opened 2015-01-13
cfd 261 upcoming must include running events

jummonk:


jummonk:
running event is upcoming

Issue #400, opened 2015-01-13
CFD-262 : BLOCKER: Dashboard page takes 34s to load after migration


zero2one:
We have a blocker for the dashboard page:

I migrated all content from D6 to D7.
The dashboard pages take 34s to load (I had to increase maximum_execution_time to test the page).

Blackfire.io shows that most time is spend on comment_node_load().
See https://blackfire.io/profiles/57c26a96-27d8-4790-9eb8-61519c6fa8e8/graph

Even 300s is not always enough:
![c4m-dashboard-error](https://cloud.githubusercontent.com/assets/133124/5723359/dafa7fd6-9b43-11e4-873a-acae0a331a8e.png)

zero2one:
Removing the activity stream pane from the group dashboard fixes the problem!

HelenaEksler:
@zero2one  I tried to load all groups from [page 1](http://capacity4more-acc.dev.amplexor.com/) And I didn't catch any problems with loading. The only thing I saw - errors when loading Javascript related as I understand to YouTube videos.
![selection_171](https://cloud.githubusercontent.com/assets/8306902/5735762/0cabde04-9bd4-11e4-9207-e7333c9f601d.png)
![selection_169](https://cloud.githubusercontent.com/assets/8306902/5735763/0df61b58-9bd4-11e4-9166-c22f1d741c87.png)

zero2one:
@HelenaEksler 
The new version is not deployed on acc yet (running the migration locally).
(acc is still the version as on the end of sprint 3).

I send you all an email with a link to a database dump of the migrated content on the current develop branch.

zero2one:
Fix for 300s+ is alre

PR #401, opened 2015-01-14
CFD-262 : BLOCKER: Fixed loading dashboard.

HelenaEksler:
#400


zero2one:
@HelenaEksler @amitaibu 
This fixes the issue but the performance of the group dashboard is still not optimal.
I will create a follow-up ticket for the next sprint.

PR #406, opened 2015-01-14
cfd 258 group events add events navigation

jummonk:


jummonk:
events navigation

zero2one:
Needs a re-roll

zero2one:
There is something wrong with the facets…

Past & Upcoming say both 6 items while the overview page says that there are 5 items per list.
Could it be that one of the events is counted twice?

The Upcoming & Past views (separate pages) both show 6 items?


c4m-events-wrong-count

zero2one:
I know I told you differently but…

Can we keep the filters also when we switch from Upcoming to Past and vice versa (like we do from overview to upcoming/past).

Issue #420, opened 2015-01-19
CFD-256 : Group Permissions : overview pages should not be accessible for restricted & private groups


HelenaEksler:
The group content overview pages:
no wiki pages page
discussions
library
events
task lists
members
Should not be accessible for non group members.
Restricted groups:
The logged in user should see them IF his email address is allowed for that group.

HelenaEksler:
@zero2one can we convert those Views pages to be under Panels? Like this we wouldn't have to develop anything, as Panels will already return an access denied if no OG context is found.

zero2one:
Implement the fix for the group dashboard (that is a panel).

We will then check & replace the views pages (overview pages) and replace them by panels.

zero2one:
QC OK > Merged.

PR #421, opened 2015-01-19
WIP: CFD-256 : Group Permissions : overview pages should not be accessible for restricted & private groups

HelenaEksler:
#420


HelenaEksler:
@jummonk Do you need this PR? Or I can close it?

jummonk:
@HelenaEksler Yes I need this PR, continuing on this one. I will close my other related PR.

amitaibu:
@jummonk My suggestion here is to move those pages to be under Panels - so if user won't have access, og_context() will return FALSE -- and the page will return a 403

jummonk:
I'm afraid that is not a solution as we meet the same issue here. The only difference is that ctools_access_menu is called instead of views_access, and the access check we add to the page using ctools is similar to the one in views_access_plugin. So same issue:
og_purl_init_provider()-> 
  og_context()->
    og_context_determine_context()->
    menu_get_item()->
      _menu_translate()->
        _menu_check_access()->
          ctools_men_access() ->
            (calling our ctools access plugin -> 
              og_context() (and now the context array is initialized and empty => access = FALSE

amitaibu:
I believe you wouldn't need to do an access check, as the page will require a context - provided by OG context. Just by the lack of having that context, the page will fail.

@HelenaEksler can you please confirm and provide the screenshots

HelenaEksler:
og-context is working for dashboard properly.
I think maybe it's a good idea to try to re-roll this?
In the second window no user is logged in.
video2

jummonk:
@amitaibu @HelenaEksler 
The test you see here is on a private group. The current user is not a member. You see that the nobelprize homepage is indeed not found, but on the events overview page (this one in panels with og_context), the views are still shown. The group menu and title is indeed missing but that is not enough. The view itself does not need the og_context to appear. In other words, panels and using og_context as context does not seem to be a solution for views within the group.
og-context-for-view-in-panel

jummonk:
@amitaibu @HelenaEksler How I can solve it. When determining the context, don't by default load the menu item and just test on current path. Because if during the custom menu access check og_context is called (as is the case in our views access plugin or ctools access plugin), then the context will seem empty. Is this an acceptable solution for you?

og_context-access-issue

amitaibu:
Yes, that's fine

jummonk:
@amitaibu See PR https://github.com/Gizra/og/pull/29

amitaibu:
@jummonk ok, until this would be commited (a simpleTest needs to be added) you can change the make file to grab your branch.

PR #425, opened 2015-01-20
Define coder version to avoid Travis failure.

HelenaEksler:


nader77:
@zero2one In which task i should log the time spent on this PR ?

zero2one:
Use CFD-270

PR #431, opened 2015-01-21
CFD-232 Allow group members with right permission to view unpublished group cont...

jummonk:
...ent.

(Implement permission, grants and node_access records.)


amitaibu:
Maybe we can use https://www.drupal.org/project/og_moderation here?

jummonk:
@amitaibu, thanks for the tip, I didn't know this one. But frankly I prefer our own custom implementation of node_access_records and grants, as more of that (realms) is coming up in the future for this project. The module you mention indeed offers among other things access to unpublished nodes. 2 reasons why I would not use that module (1) It also offers other things we don't need: access to revisioning options. I don't think we'll be needing that. @zero2one, correct me if I'm wrong. (2) As we will need other node_access realms, we would have 3 modules with a node_access_records implementation then (og_access, og_moderation and c4m_og.) 
I would prefer to limit modules with those implementations as much as possible to control the access spaghetti.

jummonk:
unpublished-wiki-pages

zero2one:
The wiki menu is only working when you are on the unpublished wiki page.

Issue #436, opened 2015-01-30
CFD-253 : Create Group : Fix the order of the fields


zero2one:
The field order should be:
* Group name.
* Path (is now shown after the group access).
* Group access
* Membership requests (only shown if you have selected Public or Restricted)
* Description
* Group Topics
* Regions & Countries
* Banner image
* Thumbnail image
* This is a subgroup of
* Related Projects
* Message to the site administrator

![c4m-group-form](https://cloud.githubusercontent.com/assets/133124/5973209/e2f8365e-a865-11e4-94ff-0934c406db73.png)

zero2one:
QC OK > Merged.

PR #443, opened 2015-02-01
CFD-253 : Create Group : Fix the order of the fields 

HelenaEksler:
#436


HelenaEksler:
Admin:
admin-c4m

Logged in user:
user-c4m

amitaibu:
Is the JS working properly on the page? - by the screenshot it looks like there's an error

HelenaEksler:
For non-admin user everything working properly. 
selection_192

For admin user there is design problem.
selection_193

Issue #440, opened 2015-01-30
CFD-279 : File Widget : Dropzone to replace an uploaded file should be larger


zero2one:
When you upload a document using the file widget, the information about that file is displayed.

When you want to replace that file you can drag-and-drop a new file.

The problem is that the space to drop the file is very small + there is no visual clue that you can drop it there.

![c4m-qp-document-dropzone](https://cloud.githubusercontent.com/assets/133124/5976406/b164d278-a88d-11e4-94ca-fa9e07d7cf96.png)

Can you enlarge the drop zone so it covers the whole file info area:

![c4m-qp-document-dropzone_2](https://cloud.githubusercontent.com/assets/133124/5976410/b5068836-a88d-11e4-90cc-858459231520.png)

zero2one:
QC OK > Merged

PR #444, opened 2015-02-01
CFD-279 : File Widget : Dropzone to replace an uploaded file should be larger

HelenaEksler:
#440


HelenaEksler:
screenshot from 2015-02-01 10 40 34

PR #449, opened 2015-02-02
ON HOLD: cfd 278 group access changing group access reindex group content

jummonk:


jummonk:
@amitaibu We have a problem here concerning access type change of a group node.
We need to 'tell' search_api to reindex the group content nodes when a group node has changed. When a group is changed from public to private, there is no problem because the Views do an access check and deny users without appropriate rights to see the group content overview. But when a group access type changes back from private to public we have an issue. The user can see the view, which is correct, but the view seems empty as the group content nodes in the search index are still tagged with the node_access value of a group content node which is part of a private group. 
I was hoping to tag the group content nodes for reindexing at entity_update (see c4m_og.module) but I'm afraid this does not occur in time. Any idea how we could solve this?

Issue #437, opened 2015-01-30
CFD-276 : Quick Post : Creation of discussion through quick post on IE9 => error


zero2one:
JavaScript Error: TypeError: Unable to get value of the property 'textAlign': object is null or undefined 

![c4m-quick-post-error-saving-discussion](https://cloud.githubusercontent.com/assets/133124/5973796/c623dd74-a86c-11e4-9fd8-9f1dac5eebc3.png)

zero2one:
QC OK > Merged

PR #450, opened 2015-02-03
CFD-276 : Quick Post : Creation of discussion through quick post caused errors for non-admin

HelenaEksler:
#437


HelenaEksler:
ck-editor doesn't cause the saving problem.
Incorrect permissions, that caused the problem, were fixed. 
Not logged in user can't create quick post. Logged in user (also non-admin) can create quick post. Works in all browsers (Chrome, Firefox, IE)

In IE:
video7

HelenaEksler:
@amitaibu Can be assigned to Peter.

Issue #452, opened 2015-02-03
Activity stream : Show more button does not work for non-logged-in users


zero2one:
The show-more button does not work if a user is not logged in.

zero2one:
QC OK > Merged

nader77:
@zero2one Number of issue in Jira ?

zero2one:
CFD-283

PR #453, opened 2015-02-04
Activity stream : Show more button for non-logged-in users.

HelenaEksler:
#452


HelenaEksler:
@amitaibu Here was the same problem, like with authenticated user. PR can be assigned to Peter.

amitaibu:
Ok, Please assign to him once green

PR #477, opened 2015-02-11
CFD-306 : Send html mail

zero2one:
### Added dependencies
* Added the mimemail module.
* Added the mailsystem module

### Integration in capacity4more
Added a custom module (c4m_mail) to do the basic configuration of the platform so it supports sending out email in the HTML format.

Added an install hook that replaces all default mail messages by a HTML version of them.


zero2one:
Ready for review

nader77:
@zero2one Installed and reviewed, all good except I got clipped message to my gmail, Is it something we can fix ? >
selection_044

selection_042

zero2one:
@nader77 Gmail does that if the mail is more then 102KB.
See: https://econsultancy.com/blog/65360-prevent-gmail-clipping-your-message-with-an-email-diet/

It is probably trigged by the css that is to big.

I will put the css on a diet ;-)

zero2one:
@nader77 I remove a lot of fat from the css in the mail.

Can you test again?

nader77:
@zero2one Sorry, Still seeing empty messages (iPhone), Clipped message with link (gmail - PC).
I can see the whole message in the mail client on a mac.

zero2one:
Did you rebuild the theme css before you run the second test?

I just did a new test and I got the message without clipping:

c4m-mail-gmail

zero2one:
The build is broken due to a .info file within the grunt-contrib-imagemin plugin.
Drupal thinks its a theme and tries to add it to the system table.

The file path is to long and does not fit within the filename field.

Issue #475, opened 2015-02-10
CFD-311 : Homepage: Add video to header about "what is C4D"


nader77:
Only for anonymous users:
Add a video link to the header of the website.
The link opens a popup window in an overlay that plays a youtube video

zero2one:
@nader77 The video needs to be in the header of the site not in the panel.

The video (for not-logged-in users) is not within the panel but within the header of the site.
You will need to create a block for it and put it there using context.

zero2one:
QC OK > Merged.

PR #488, opened 2015-02-12
CFD-311 : Homepage: Add video to header about "what is C4D"

nader77:
#475


nader77:
@HelenaEksler Added pause on video hide script, Please review.

HelenaEksler:
@nader77 Looks fine. 

@zero2one Banners that should be in this block - will be added when we get the images, in the following PR.

capacity4more/modules/c4m/features/c4m_features_homepage/plugins/content_types/video_embed/video-embed.tpl.php
```diff
@@ -0,0 +1,31 @@
+<script type="text/javascript">
```
zero2one: Can we put the .javascript in a separate javascript file and include in the page trough drupal_add_js() in the c4m_features_homepage_video_embed_content_type_render() callback?

The javascript should be loaded in the footer.

nader77: @zero2one I need this piece of js in the tpl, because i need to pass the id of the youtube video to it. as you can [see here](https://github.com/capacity4dev/capacity4more/pull/488/files#diff-74bfb8652085e4ea7ad3aa06365a8575R6), Can i pass it to a js file ?

HelenaEksler: @nader77 You can pass it in the same way like in angular - through the settings.

nader77:
@zero2one Fixed, Please review.

zero2one:
@nader77 / @HelenaEksler : Needs a follow up theming task.

Issue #498, opened 2015-02-17
CFD-293 Homepage: Activity stream based on the filter (Logged-in users)


HelenaEksler:

zero2one:
QC OK > Merged.

PR #499, opened 2015-02-17
CFD-293 Homepage: Activity stream based on the filter (Logged-in users)

HelenaEksler:
#498


capacity4more/modules/c4m/message/c4m_message/c4m_message.module
```diff
@@ -849,8 +849,59 @@ function c4m_message_preprocess_node__event(&$variables) {
...
+        if (!empty($interests)) {
+          $query->fieldCondition('c4m_related_topic', 'target_id', $interests, 'IN');
+        }
+        $result = $query->execute();
+        $groups += !empty($result['node']) ? array_keys($result['node']) : array();
+      }
+    }
+    else {
+      // Get all groups user has access to.
+      $query = new EntityFieldQuery();
```
amitaibu: this looks wrong - we can't get *all* the groups in the system

amitaibu: at the minimum there should be a limit

nader77:
@HelenaEksler As you are working on the activity stream, I found a bug when the user has no groups, he will have an error when automatically refreshing the activity stream.
Use user "president".
Thanks.

HelenaEksler:
selection_230
selection_231

zero2one:
@HelenaEksler when I filter by "My Groups" I still see Article activity in the list.

Articles don't belong to a group and should not be listed in the stream…

c4m-activity-filter-groups

HelenaEksler:
@zero2one It's strange - because I don't see articles when filter is set to My groups.
capacity4more

Maybe you should try to re-install.

zero2one:
@HelenaEksler 

I got also a fatal error in the theming layer when I:

* Log in as "president"
* Set his interests to "earth"
* Go to the home page
* filter by interests => Fatal error.

I'm rebuilding the platform to be sure that nothing went wrong…
I'll keep you posted

HelenaEksler:
@zero2one  Same thing here - don't get any errors
selection_232

selection_233

zero2one:
@HelenaEksler After re-install:
* Filter by my groups works
* No PHP errors.

zero2one:
QC OK > Merged.

PR #501, opened 2015-02-17
CFD-298 : Homepage : Upcoming events (Logged-in users) 

HelenaEksler:
#494


HelenaEksler:
@nader77 I made some changes here - so could you review this, please.

HelenaEksler:
selection_224

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
```diff
@@ -219,38 +219,110 @@ function c4m_features_og_events_entity_info_alter(&$entity_info) {
...
+
+        $groups += !empty($result['node']) ? array_keys($result['node']) : array();
+      }
+      break;
+    case 'groups':
+      $groups = og_get_groups_by_user();
+      $groups = !empty($groups['node']) ? $groups['node'] : array('no_groups');
+      break;
+    default:
+      break;
```
nader77: Shouldn't we have a default case ?

HelenaEksler: By default array is empty (for logged in users) or contains public groups (for anonymous user) and this array is defined before switch. https://github.com/capacity4dev/capacity4more/pull/501#discussion-diff-24885378R246 and https://github.com/capacity4dev/capacity4more/pull/501#discussion-diff-24885378R242

nader77: Yea i know, I thought maybe we can change the code to have this happen in the default of the switch.
Anyway. Looks good otherwise, i will assign to Peter

nader77:
@zero2one Ready for your review.

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
```diff
@@ -219,38 +219,110 @@ function c4m_features_og_events_entity_info_alter(&$entity_info) {
...
+  if ($user->uid) {
+    // Check if content filter is applied.
+    $user_wrapper = entity_metadata_wrapper('user', $user->uid);
+    $interests = $user_wrapper->c4m_related_topic->raw();
+    $filter = !empty($_GET['filter']) ? $_GET['filter'] : '';
+    $groups = array();
+    $range = 3;
+  }
+  else {
+    $groups = c4m_og_get_groups_by_access_type('public');
```
zero2one: It's not only the events of public groups a user has access to.

It can be that the user has an email address domain that gives him access to all restricted groups who granted access to this domain.

### Solution
Get 3 events that:

* Startdate >= midnight tomorrow
* Filtered by the homepage filter:
  * Any
  * By interests (topics)
  * Within my groups
* Show only events I have access to (use the node_access system).

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
```diff
@@ -219,38 +219,110 @@ function c4m_features_og_events_entity_info_alter(&$entity_info) {
  * 1 Event for anonymous user, 3 for logged in.
  *
  * @return string
- *  HTML of rendered upcoming events.
+ *  HTML of rendered upcoming events, FALSE if there's no events.
  */
 function c4m_features_og_events_upcoming_events() {
```
zero2one: Split this function in 2:

* One to get the events when user is not logged in (1 event).
* One to get the events when user is logged in (3 events filtered by the homepage filter).

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
```diff
@@ -219,38 +219,110 @@ function c4m_features_og_events_entity_info_alter(&$entity_info) {
...
+  );
+  $events .= theme_html_tag($title);
+
+  // Render events.
+  foreach ($result['node'] as $event) {
+    $node = node_view(node_load($event->nid), 'upcoming_event');
+    $events .= drupal_render($node);
+  }
+
+  // Render "see more" link.
```
zero2one: Use the l() function instead?

capacity4more/modules/c4m/features_og/c4m_features_og_events/plugins/content_types/upcoming_events/upcoming_events.inc
```diff
@@ -15,7 +15,7 @@ $plugin = array(
 function c4m_features_og_events_upcoming_events_content_type_render($subtype, $conf, $args, $context) {
```
zero2one: Create 2 different renderes:

1. Upcoming event for non-logged in users.
2. Upcoming events for logged in users.

capacity4more/modules/c4m/og/c4m_og/c4m_og.module
```diff
@@ -337,6 +337,7 @@ function c4m_og_node_access_records_alter(&$grants, $node) {
  *  Entity IDs of the public groups.
  */
 function c4m_og_get_groups_by_access_type($access) {
```
zero2one: The title of this function is misleading: is suggests that you can get all groups by their access type (public, restricted, private).

Thats not the case.

Why should we pass the public 0/1 parameter, can't we rely on the node_access system to get only the groups the user has access to?

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
```diff
@@ -219,38 +219,110 @@ function c4m_features_og_events_entity_info_alter(&$entity_info) {
...
+  $link['element'] = array(
+    '#tag'        => 'a',
+    '#value'      => t('See More ') . '<i class="fa fa-chevron-right"></i>',
+    '#attributes' => array(
+      'href' => url('events'),
```
zero2one: The proper link is "global-events"

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
HelenaEksler: @zero2one Are you sure we should split plugin into 2? These would be almost the same plugins and functions. I think we shouldn't have duplicated code.

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
HelenaEksler: Because I need to put  ``<i class="fa fa-chevron-right"></i>`` right after the link.

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
zero2one: This code is way to complex: 1 function is:

* Defining if we need 1 or 3 items
* If the user is logged in or not
* Query for the items
* Checking what items to load based on the filter
* Render the output
* ...

This is a recipe for bugs (way to much paths in 1 function).

I would suggest:
2 functions (1 event, 3 events for logged in users).

Each function should call other functions:

* Call a function to get the nodes.
* Call a theme function that renders the nodes in an overview with a link to the events overview.

The function that gets the events for logged in users should:
* call a helper function to get the active filter.
* call the function to collect the nodes to display with the filter param.

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
zero2one: I think it's event better to put the rendering of the output in a template.
Then you don't have to add the title & link in render arrays…

And gives the advantage we can override the template later if necessary.

We should avoid 1 function doing to much.
We should put rendering of output in theme functions and/or templates.

HelenaEksler:
@zero2one Review, please.

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
```diff
@@ -213,44 +213,135 @@ function c4m_features_og_events_entity_info_alter(&$entity_info) {
...
-function c4m_features_og_events_upcoming_events() {
-  global $user;
+function c4m_features_og_events_upcoming_events_loggedin() {
   $events = '';
 
+  // Get content filter.
+  $filter = !empty($_GET['filter']) ? $_GET['filter'] : '';
+  $range = 3;
+
+  $groups = c4m_features_og_events_get_groups_by_filter($filter);
```
zero2one: The filtering system is wrong.

The events should **not be filtered by the groups** that have topics the user is interested in.

The events should **be filtered by the events** that have topics the user is interested in.

capacity4more/modules/c4m/features_og/c4m_features_og_events/c4m_features_og_events.module
```diff
@@ -213,44 +213,135 @@ function c4m_features_og_events_entity_info_alter(&$entity_info) {
...
+ * Get rendered events for loggedin user.
  *
  * @return string
- *  HTML of rendered upcoming events.
+ *  HTML of rendered upcoming events, FALSE if there's no events.
  */
-function c4m_features_og_events_upcoming_events() {
-  global $user;
+function c4m_features_og_events_upcoming_events_loggedin() {
   $events = '';
```
zero2one: Remove, variable is not used as it is replaced by 

```
$events = c4m_features_og_events_upcoming_events_get_events($range, $groups);
```

HelenaEksler:
@zero2one 
Due to wiki events should be filterred by groups, that have topic the user is interested in.

selection_227

selection_226

zero2one:
@HelenaEksler Thats an error in in the specs: it should be the events that have topics that match the topics of interests of the user. The "interests" filter is not based on groups, only on topics.

zero2one:
I corrected the specs:

* https://wiki.amplexor.com/confluence/display/C4DNG/Home+page+-+Authenticated+user+-+No+member
* https://wiki.amplexor.com/confluence/display/C4DNG/Home+page+-+Authenticated+user+-+Member

HelenaEksler:
@nader77 Can you review it?

nader77:
@HelenaEksler Change the ``c4m_og_get_groups_by_access_type($access)`` function name or make it get only public groups and change the name (As you see fit), other than that, all good assign to peter when you fix.

HelenaEksler:
@zero2one PR is ready.

zero2one:
QC OK > Merged.

Issue #511, opened 2015-02-23
CFD-298 : Homepage : Upcoming events - Theming


zero2one:
Follow up of:

* #501
* #494 

![c4m-homepage-events](https://cloud.githubusercontent.com/assets/133124/6328536/cf22978a-bb68-11e4-8cae-a0c1ac0eaf65.png)

zero2one:
theming ok > merged.

PR #512, opened 2015-02-23
CFD-298 : Homepage : Upcoming events - Theming

HelenaEksler:
#511


HelenaEksler:
@zero2one Can I leave event date under the label and to use this icon?
selection_234

zero2one:
@HelenaEksler yes: looking good!

Issue #525, opened 2015-02-26
CFD-319 : Homepage : group owner is missing in the my groups & suggested groups list


nader77:
The owner of the group should be listed in the group overview blocks.

zero2one:
QC OK > Merged.

PR #527, opened 2015-02-26
CFD-319 : Homepage : group owner is missing in the my groups & suggested groups list

nader77:
#525 
This was added and styled, apparently not exported correctly.

* Fixed the export.
* Added a test to make sure it's always there.

selection_072


nader77:
@zero2one Could you please review this.

PR #531, opened 2015-03-17
CFD-367 People : Global people list overview

MPParsley:
This feature branch also includes CFD-368 People : Global people table overview.


MPParsley:
global_people_list

Issue #535, opened 2015-03-22
CFD-369: In the picture : create new content type


nader77:
Create a new content type "inthe_picture":

New module:
* Module name: c4m_content_inthe_picture
* Content type human name: In the Picture 
* Content type machine name: inthe_picture
* Pathauto path : inthe_picture/[node:title]

Fields:
* *Image* (mandatory) : a wide image (similar to Voices & views and Groups) to use as the background of the "in the picture" block.
  Use the c4m_media field.
* *Title* (mandatory) : a title to use in the display of the block.
  Use the node title field.
* *Body* (mandatory) : some plain text to use in the display.
  Use the c4m_body field.
* *Link* (mandatory) : a link to the internal path or external link.
  Use the c4m_link field.
* *Page* (mandatory) : on what global overview page should this block been shown.
  Possible pages:
    * Voices & Views
    * Groups
    * Projects
    * Events
    * People

PR #538, opened 2015-03-23
CFD-369: In the picture : create new content type

nader77:
#535 

Added the content type and the module.
selection_066

selection_067

@zero2one There's a text field called ``c4m_page`` which is a text field, I want a select list field for this content type, Right ?
If yes, What should i call it ? (Now it's ``field_page``),


nader77:
Update: Added demo migration to the content type:

* Nodes created in the demo migration:

selection_068

* Node edit form (Filled fields):
selection_069

zero2one:
@nader77 
* Set the label of that field to: "Overview page"
* Set the description of the field to: "Select the overview page where this in the picture item should be shown."

The multiple checkboxes are ok.

nader77:
@zero2one Field machine name ``c4m_overview_page`` is ok ?

zero2one:
@nader77 yes

HelenaEksler:
@nader77 Is it ready? If so - assign to Peter.

PR #543, opened 2015-03-23
CFD-360 Voices & Views overview

MPParsley:
Overview page of all Voices & Views:
- A search view
- Only published Voices & Views
- Order by post date (desc) by default
- Showing teasers
- Paged, 10 items per page.

voices views overview


MPParsley:
Added filter on topics.
Show title on top.

voices views

MPParsley:
Demo articles

voices views demo content

zero2one:
@MPParsley 

Voices & Views overview:
* There is a missing list image for the "article/millennium-windmill-development-and-modern-engineering" article.

Groups overview : 
* Filter : the filter should show "My Groups" filter only when the user has groups (do not count projects).

MPParsley:
Fixed the image and added API calls to get groups for a user and that make the difference between projects & groups.

PR #557, opened 2015-03-28
CFD-361 voices & views detail

MPParsley:
article detail


MPParsley:
Shouldn 't the title be displayed at the top, above the banner?

Issue #568, opened 2015-03-31
CFD-412: Group Approval Workflow : Update migration


solker:
We need to update the migration so the groups get the proper status:

* Groups that are published > Published
* Groups that are unpublished > Draft

Make sure that no notifications are send during the migration.

https://issuetracker.amplexor.com/jira/browse/CFD-412

zero2one:
Migration code is in : https://github.com/capacity4dev/c4d_import

solker:
@zero2one what about a dump of data for test?

solker:
@zero2one demo content with published status already have Published status, and draft content also have Draft status. Content with unpublished status not exists in demo content. Therefore I can change behavior of migrate module only for unpublished items that should be Draft.

zero2one:
It's not about the demo content it's about the migration of the real data.

We will take care of it.

PR #550, opened 2015-03-25
CFD-374: Group approval workflow - update form

solker:
#544

PR #570, opened 2015-04-01
CFD-412: Group Approval Workflow : Update migration

solker:
#568

* First commit for https://github.com/capacity4dev/capacity4more/pull/550/files#r27391678


solker:
* Second commit to change default behavior for group status field in migration module. 

@HelenaEksler ready for code review and merge.

HelenaEksler:
@zero2one I think this solution will solve the migration problem - migrated groups by default will be with status "published". So you can merge it.

Issue #572, opened 2015-04-02
CFD-410: Activity stream: Do not show unpublished content


solker:
According to the specs an article update shouldn't be displayed in the activity stream.
More importantly it is now possible for an anonymous user to see activity on unpublished content (see screenshot).

https://wiki.amplexor.com/confluence/display/C4DNG/Activity+stream

PR #573, opened 2015-04-02
CFD-410: Activity stream: Do not show unpublished content

solker:
#572


capacity4more/modules/c4m/message/c4m_message/c4m_message.module
```diff
@@ -109,7 +109,6 @@ function c4m_message_crud_message($entity_type, $entity, $op) {
     'photoalbum',
     'document',
     'wiki_page',
-    'article',
```
HelenaEksler: @solker Remove articles only for operation "insert".

HelenaEksler:
@solker Check why test are red.

capacity4more/modules/c4m/message/c4m_message/c4m_message.module
solker: @HelenaEksler check now

Issue #595, opened 2015-04-06
CFD-388 : Group management Categories : Reorder Category terms.


nader77:
Add the possibility to change the order of the category terms.
Use the Drupal drag-n-drop interface element for this.
Add a link above each Category list to change the order of the terms back to Alphabetically.

Module: ``c4m_features_og_manage``
Checkout from: ``feature/CFD-384-GMC_overview_screen``

nader77:
@YuriGizra ^^

PR #581, opened 2015-04-05
CFD-384 : Group management Categories : Overview screen.

nader77:
#571

PR #600, opened 2015-04-06
WIP. CFD-388 : Group management Categories : Reorder Category terms.

YuriGizra:
#595


nader77:
Merging this to PR #581 because we need the fixes here and it's from the forked repository.

Issue #608, opened 2015-04-08
CFD-332 : Cannot select past date in QP event.


nader77:

zero2one:
QC OK > Merged

PR #609, opened 2015-04-08
CFD-332 : Cannot select past date in QP event.

nader77:
#608 

Selecting past date:
selection_095

Created event with past dates:
selection_096


nader77:
@HelenaEksler This can be reviewed and assign to maarten so maybe he can merge it.
Thanks.

PR #624, opened 2015-04-08
CFD-351 : QP should not appear for a public visitor.

nader77:
#623 

* Hidden QP form for anonymous users.
* Style changes to activity stream to fit without the QP form.

Logged in user:
selection_097

Anonymous user: 
selection_098


nader77:
@HelenaEksler PR is ready for code review.

HelenaEksler:
@nader77 Looks fine. The only one question - don't you want not to load the whole information about fields and so on? You've just hided the theme part, but all JS settings are still loading.

amitaibu:
As we'll probably need to tackle QP to be lazy loaded, lets do an extra effort in not loading QP at all if not needed.

nader77:
@HelenaEksler @amitaibu Fixed, Only activity stream settings are loaded for anonymous users.

HelenaEksler:
@nader77 Looks ok. As soon as tests are green assign to Peter.

nader77:
PR is ready for review.

zero2one:
@nader77 I got this question while installing capacity4more. What is the proper Angular version we should use and can we fix this so no question pops up (will break Travis-CI builds).

c4m-what-bower-version

zero2one:
Same for angular-animate
screen shot 2015-04-20 at 08 39 28

angular-touch
screen shot 2015-04-20 at 08 39 36

Issue #626, opened 2015-04-12
CFD-340 error saving on QP document creation


HelenaEksler:
When creating a document through the QP, I dragged a file, then added a title and completed the form with all tags - but when I press save, I get a message below the save button that says "error saving" - this also generates complete loss of data except for the file that is still attached, as can be seen on the screenshot attached.
This is not the first time it happens during testing. I even tried to complete the information a second time and got the same bug, with only the file remaining and no data, but an error message. Can you please investigate? Thanks

PR #629, opened 2015-04-12
CFD-340 error saving on QP document creation

HelenaEksler:
#626


HelenaEksler:
video151

nader77:
PR is ready.

nader77:
@MPParsley Fixed the permissions, Had to use 'edit term' since there's no 'create term'.
Had to add permissions to members upon install, there's a PR to handle the OG content permissions #632 which will complete this one and set us on the right path to handle CFD-432.

PR #636, opened 2015-04-14
CFD-342 - Always max-width of 100%

soniCaH:


zero2one:
Looks obsolete since we have a teaser display mode in place with a reduced image size.

zero2one:
Already fixed in other Pull request.

PR #645, opened 2015-04-16
CFD-322: group icons in IE and safari

SebastienVanRintel:
Set x and y for background-size on group icons for better displaying in IE and safari
image

Fixed issue where checkboxes weren't shown in IE on group page
image


soniCaH:
Temporary disabled SVG's as a solution needs to be found to ensure icons appear the same cross-browser.

Safari on windows for example does not use the background-size properly, causing icons to disappear.

PR #652, opened 2015-04-20
CFD-330 : Display full name in messages (activity stream)

soniCaH:

PR #632, opened 2015-04-13
CFD-347 : Cannot edit my own discussion.

YuriGizra:
[CFD-347](https://issuetracker.amplexor.com/jira/browse/CFD-347)
#631

PR #655, opened 2015-04-21
CFD-330 : Display full name in messages (activity stream) - Fix tests

nader77:
Follow up on #652

### DIFF
--- a/capacity4more/behat/composer.lock
+++ b/capacity4more/behat/composer.lock
@@ -761,26 +761,27 @@
         },
         {
             "name": "symfony/browser-kit",
-            "version": "v2.5.6",
+            "version": "v2.6.6",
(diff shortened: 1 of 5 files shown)


nader77:
The tests are fixed, But i think it will fail again because it will exceed the 50 min limit in travis,
Gonna wait for PR #632 to be merged and divide the ``javascript`` tests.

This is on HOLD.

zero2one:
We are, from now on, using the payed version of Travis:
* Tests can run up-to 120mins
* We can run 5 workers at the same time.

There is currently an issue on the side of Travis (queue problem with Linux builds), but when they fix it we can add more tests without worrying about timeout and splitting them in smaller chunks.

nader77:
@zero2one PR is ready.

PR #656, opened 2015-04-21
CFD-433 Request a group : The theme for this page should be kapablo, not seven

HelenaEksler:
https://issuetracker.amplexor.com/jira/browse/CFD-433

### DIFF
--- a/capacity4more/modules/c4m/content/c4m_content_group/c4m_content_group.module
+++ b/capacity4more/modules/c4m/content/c4m_content_group/c4m_content_group.module
@@ -295,17 +295,11 @@ function c4m_content_group_submit_redirect(&$form, &$form_state) {
 }
 
 /**
- * Implements hook_menu_alter().
+ * Implements hook_admin_paths().
  *
- * Set Kapablo theme for group creation page.
+ * @see og_ui_admin_paths().
+ * Set the 'node/add/group' path as not admin.
  */
-function c4m_content_group_menu_alter(&$items) {
-  $items['node/add/group']['theme callback'] = 'c4m_content_group_theme_callback';
-}
-
-/**
- * Returns Kapablo theme name.
- */
-function c4m_content_group_theme_callback() {
-  return 'kapablo';
+function c4m_content_group_admin_paths() {
+  return array('node/*/group' => FALSE);
 }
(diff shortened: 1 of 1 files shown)


HelenaEksler:
selection_308
selection_309

HelenaEksler:
@nader77 PR is ready.

capacity4more/modules/c4m/content/c4m_content_group/c4m_content_group.module
```diff
@@ -293,3 +293,19 @@ function c4m_content_group_submit_redirect(&$form, &$form_state) {
   // Redirect a user to homepage.
   $form_state['redirect'] = url('', array('absolute' => TRUE));
 }
+
+/**
+ * Implements hook_custom_theme().
+ *
+ * Set Kapablo theme for group creation page.
+ */
+function c4m_content_group_menu_alter(&$items) {
```
nader77: This is not custom_theme, Could we do it like:
```
/**
 * Implements hook_custom_theme
 */
function MYMODULE_custom_theme() {
  // Set path specific theme.
  if (current_path() == 'node/add/group') {
    return variable_get('custom_theme', 'theme_name');
  }
}
```

@amitaibu Which one is better ?

amitaibu: hook_custom_theme() seems more straightforward

zero2one: I don't believe this is the solution:

* The **settings for the Admin theme** already state that all content create/edit/delete forms **should be rendered in the Default theme (Kapablo)**.
* This **works for all node types** (including projects) **except Groups**.
* There is probably an overwrite already of this behaviour that breaks the proper theme for groups.
* The **group edit form does use the Kapablo theme**, it's only the create that has a problem.

capacity4more/modules/c4m/content/c4m_content_group/c4m_content_group.module
```diff
@@ -293,3 +293,19 @@ function c4m_content_group_submit_redirect(&$form, &$form_state) {
   // Redirect a user to homepage.
   $form_state['redirect'] = url('', array('absolute' => TRUE));
 }
+
+/**
+ * Implements hook_menu_alter().
+ *
+ * Set Kapablo theme for group creation page.
```
zero2one: I don't believe this is the solution:

The settings for the Admin theme already state that all content create/edit/delete forms should be rendered in the Default theme (Kapablo).
This works for all node types (including projects) except Groups.
There is probably an overwrite already of this behaviour that breaks the proper theme for groups.
The group edit form does use the Kapablo theme, it's only the create that has a problem.

PR #661, opened 2015-04-22
CFD-331 Add 'Add an event' button in the header of the event overview

SebastienVanRintel:
Added 'Add an event' button in the header of the event overview page-view and removed the button from the upcomming-events block
Updated 'Add an event' button css for better layout
Rearanged weights of group events overview blocks so they come under the page content

image


MPParsley:
no events

PR #675, opened 2015-04-28
CFD-440 - Member data - country of residence not shown

soniCaH:
Reversed logic was hiding the field if a correct country was entered, and showing the value if - choose country or region - (default if not migrated correctly) was selected.

screen shot 2015-04-28 at 13 36 31


capacity4more/modules/c4m/user/c4m_user/includes/c4m_user.theme.inc
```diff
@@ -233,7 +233,7 @@ function theme_c4m_user_country(array $field) {
   if (empty($country)) {
     return NULL;
   }
-  if ('- choose country or region -' !== drupal_strtolower($country)) {
```
zero2one: Use `t('- choose country or region -')` to avoid breaking functionality when the interface becomes multilingual…

PR #680, opened 2015-04-28
CFD-487 Fixed css layout on event date/location

SebastienVanRintel:
image


MPParsley:
The date line should also start on a new line if no organiser is available.

schermafdruk 2015-04-28 16 44 10

SebastienVanRintel:
Event without a group:
image

More padding on the event type icons in the event form
image

PR #670, opened 2015-04-27
CFD-429 : Quickpost to full form -> file upload fail

nader77:
https://issuetracker.amplexor.com/jira/browse/CFD-429

PR #685, opened 2015-04-29
CFD-454 create in full form generates a bug that clears data from the QP

HelenaEksler:
https://issuetracker.amplexor.com/jira/browse/CFD-454

Follows up #670

### DIFF
--- a/capacity4more/modules/c4m/restful/c4m_restful_quick_post/components/c4m-app/src/controllers/main.js
+++ b/capacity4more/modules/c4m/restful/c4m_restful_quick_post/components/c4m-app/src/controllers/main.js
@@ -321,10 +321,6 @@ angular.module('c4mApp')
       // Check for required fields.
       var errors = Request.checkRequired(submitData, resource, resourceFields);
 
-      // Check the type of the submit.
-      // Make node unpublished if requested to create in full form.
-      submitData.status = type == 'full_form' ? 0 : 1;
-
       // Cancel submit and display errors if we have errors.
       if (Object.keys(errors).length && type == 'quick_post') {
         angular.forEach(errors, function(value, field) {


HelenaEksler:
video18

HelenaEksler:
@nader77 Review, please.

capacity4more/modules/c4m/restful/c4m_restful_quick_post/components/c4m-app/src/controllers/main.js
```diff
@@ -321,10 +321,6 @@ angular.module('c4mApp')
       // Check for required fields.
       var errors = Request.checkRequired(submitData, resource, resourceFields);
 
-      // Check the type of the submit.
-      // Make node unpublished if requested to create in full form.
-      submitData.status = type == 'full_form' ? 0 : 1;
```
nader77: We still need to make the node unpublished if we go to full form to continue creating it.

HelenaEksler: Since we removed tab Published/unpublished - we can't make the node published - so it should be or published from the very beginning - or we should give to user an option to make it published after he finishes creation.

nader77: The important thing is that other users won't see unfinished posts (Started editing in full form and left without finishing), We have to find a way to do that.

HelenaEksler: I'll try to figure it out.

HelenaEksler:
@zero2one 
As soon as "Save as draft/Publish" functionality provided - PR is not longer actual. QP functionality working as expected.

soniCaH:
CFD-429 was merged in develop

PR #693, opened 2015-04-30
CFD-495 Removed the sort filter from all the non-events views

SebastienVanRintel:
Removed the sort filter from all the non-events views

image


SebastienVanRintel:
Updated if else structure to be in line with the druapkl coding standards https://www.drupal.org/coding-standards

PR #702, opened 2015-05-04
CFD-482 members overview remove breadcrumbs

SebastienVanRintel:


MPParsley:
The breadcrumb is also set by facet api (see setBreadcrumb in url_processor_standard.inc).

PR #705, opened 2015-05-05
CFD-458 : Make title as link on the carousel 

solker:
https://issuetracker.amplexor.com/jira/browse/CFD-458

> We’ll have to add the link only on the title and make it visually clear it is a link.

selection_020


solker:
@MPParsley ready for review.

PR #706, opened 2015-05-05
CFD-501 Quick Post: disabling features does not disable the option to create related content trough QP

HelenaEksler:
https://issuetracker.amplexor.com/jira/browse/CFD-501


nader77:
@HelenaEksler Merged the "CFD-405 : Manage Group features : Quick post" to here.

HelenaEksler:
@MPParsley @zero2one Seems like one who merged Nader's PR before has merged it into wrong branch. 
So I fixed here permissions to see the form. Since only group admins can access the page and we can't use OG-permission for access in the hook_variable_group_info() and there is no option to add access_calback - I'm using simple content creation permission.
Also fixed small bug in the group manage page. If group doesn't have thumbnail image - we don't see error  any more.

selection_321
selection_322

PR #667, opened 2015-04-27
CFD-357 : Group Categories not transferred as expected

nader77:

PR #709, opened 2015-05-06
CFD-236 : Quick Post form : Don't load the forms as long as no button is clicked

nader77:


nader77:
Update: Lazy loading is implemented, The fields info is fetched only when clicking on one of the buttons.
The proccess takes half a second, So i added a spinner, But it's not spinning, Still need to check it out.

nader77:
Note: This PR has #706 merged into it.
This should not be merged before that one is merged.
And it's preferable not to merge before PR #667 is merged as well.
I will keep this WIP in the mean time.

nader77:
Lazy loading:
lazy_loading

nader77:
Update: Fixed QP permission, Not showing QP form at all when user does NOT have permissions to create an entity in the group. (It was showing but could not post).

nader77:
Update: Added "isaacnewton" as a group member because the user needs to see the QP form in the group dashboard and with the new settings users who can't post, Can't see the QP form at all.

nader77:
Update: All blocking branches has been merged to Develop and this is ready, Except one small design issue.

nader77:
@zero2one @MPParsley  The ``fontAwesome`` Files in the theme are creating conflicts, not working properly and they are outdated.
They not allowing ``CSS3`` animation in the website at all.
I deleted all the files and called the font-awesome CDN file, Is there a problem with that ?
Do you prefer that i download it via bower or the make file on each install ? and reference it from the libraries folder ?
Thanks.

nader77:
Spinner with the ``fontAwesome`` CDN file:
loader

nader77:
@MPParsley @zero2one ^^

soniCaH:
I'd prefer bower/make instead of calling the CDN.
We are also overriding some glyphicons (Bootstrap JS is setting the classes), but for consistency we want to display the FA icons instead.
Currently we are solving this by overriding the glyphicon classes
```
.glyphicon {
  @extend .fa;

  &.glyphicon-chevron-left {
    @extend .fa-chevron-left;
  }

  &.glyphicon-chevron-right {
    @extend .fa-chevron-right;
  }

  &.glyphicon-remove {
    @extend .fa-times;
  }

  &.glyphicon-warning-sign {
    @extend .fa-exclamation-triangle;
  }

  &.glyphicon-ok {
    @extend .fa-check;
  }
}
```

Since we have the source available, we can extend the classes, but I assume this won't be available in your method?

nader77:
@soniCaH Yes ofcourse, The bower/make option is the best one, i will make sure it's like that.

Here i see that you are overriding the ``glyphicon`` classes, which come from bootstrap, The classes of ``fontAweosme`` will be there when we only import the file, So i'm pretty sure it will work.

You can checkout this branch now and check if it's working (Now it's with CDN),
Or tell me where i can check it.

Thanks.

nader77:
@soniCaH Ohh yea, you are right, Can't extend classes, The grunt will not be able to build the css.

Is it possible to give up the extending ?

nader77:
Update: Added the ``fontAwesome`` library to the make file.
Fixed categories empty checking to work with Lazy loading.

zero2one:
@nader77  needs a re-roll to fix the broken build

nader77:
i've merged branch develop, this PR needs some questions answered ^^

soniCaH:
@nader77 OK. We might need to override some templates in the future that spit out glyphicons by default (if they are included in the final designs).

nader77:
@zero2one PR is ready for code review.

PR #660, opened 2015-04-22
CFD-432 Implement OG permissions to the QP.

HelenaEksler:
https://issuetracker.amplexor.com/jira/browse/CFD-432

PR #710, opened 2015-05-07
CFD-472 : cannot upload document from the library to attach to another content

solker:
https://issuetracker.amplexor.com/jira/browse/CFD-472

Permissions fixed in #660


HelenaEksler:
@zero2one @MPParsley 

I've found out, that the bug is caused by JQuery. Also I've found the solution.
We should put this code in the begginning of the ``Sizzle`` function inside ``/capacity4more/modules/contrib/jquery_update/replace/jquery/1.10/jquery.js`` file. 
```php
    try{
      document === document; //may cause permission denied
    }
    catch(err){
      document = window.document; //resets document, and no more permission denied errors.
    }
```

It seems that the problem is solved in the 2.0.3 version http://bugs.jquery.com/ticket/14535
So, how do you think, what is the best solution - try to override this object (actually it is the jQuery.find function) or try to use later version of the jQuery.

PR #721, opened 2015-05-08
CFD-533 : Coder review does not validate .module & .info files

zero2one:
The coder review does not include the specific Drupal extensions.

The following should be covered:
* .php
* .module
* .inc
* .install
* .test
* .profile
* .theme
* .js
* .css
* .info
* .txt

Enabling these requires fixes on existing code.


zero2one:
@SebastienVanRintel can you do a re-roll.
I will merge this as soon as all is back to green…

PR #738, opened 2015-05-13
CFD-345 : topic selection is not limited to 3 topics

nader77:
https://issuetracker.amplexor.com/jira/browse/CFD-345

This fixes Diane's comment.
The function was not stopping in the right place.
The return was inside a ``forEach`` which just stopped the loop instead of stopping the function.

Fixed:
check-limit-2015-05-13_11


nader77:
@HelenaEksler PR is ready.

HelenaEksler:
@nader77 Good. When tests are green - assign to Amplexor and move to QC in Jira

PR #746, opened 2015-05-20
CFD-589 - Migrate all url aliases for each node

soniCaH:


zero2one:
@soniCaH  needs a re-roll to fix the broken build

soniCaH:
Re-rolling

zero2one:
We need to change the migration process & alias settings.

The redirect module creates an redirect only if aliases are deleted on update.

#### We need to change configuration
* Set the variable *pathauto_update_action* to *2* (Delete old alias).
* The redirect module will create automatically a redirect when an alias is deleted.

#### We need to change the migration
The migration needs to be changed:
* All old URL aliases should be added as a redirect entry in the redirect table.

PR #747, opened 2015-05-20
CFD-523 add organisation icon to user fullname display

SebastienVanRintel:
image


zero2one:
@SebastienVanRintel  needs a re-roll to fix the broken build

capacity4more/modules/c4m/content/c4m_content_organisation/c4m_content_organisation.install
```diff
@@ -0,0 +1,47 @@
...
+    case 'c4m_domain':
+      $columns = array(
+        'domain' => array(
+          'type'      => 'varchar',
+          'length'    => 255,
+          'not null'  => TRUE,
+        ),
+        'icon' => array(
+          'type'      => 'int',
+          'length'    => 4,
```
zero2one: Length is only for char, varchar and text fields, size is for int field types.

Remove from definition.
Do not add size, the int will be by default the correct value.

zero2one:
Refactored node_access + refactored how the organisations are displayed in group creation form.

c4m-request-group

PR #748, opened 2015-05-20
CFD-590-dont_show_big_document_file_icon

SebastienVanRintel:
image

image


zero2one:
@SebastienVanRintel needs a re-roll to fix the broken build

PR #750, opened 2015-05-21
CFD-604 : fix failing tests on travis ci

zero2one:
Fix failing build tests.


zero2one:
## Problem
The Travis-CI builds started failing on Wednesday: "Segmentation fault" when a Drush command was run.
Even builds where no coding was changed (only config) the Drupal build failed.

We first thought that it had something to do with the problems Travis-CI had, but that was not the problem.


## Root Cause
We use node.js to install dependencies for the javascript that powers the Quick post, Field widgets & activity stream. It is also used to combine multiple javascript files into 1 big file.

Node installs all kind of modules that it needs to perform all these actions. It installs these modules into a node_modules folder within the c4m_restful_quick_post Drupal module.
Node modules install automatically dependencies within a nodes_module folder within their install path, modules within there install their dependencies, etc…

One of the depending modules is the karma module. That module was updated to a newer version. In that version are 2 files with the .info extension.

.info files are used within Drupal to inform the platform about Drupal modules. Drupal detects the, non-Drupal, .info files within the node_modules and throws an error. 
*This breaks the Drupal build on Travis*.

## Solution
The .info files within the karma modules don't contain data that is needed.

[This blog post|http://drupal.stackexchange.com/questions/126880/how-do-i-prevent-drupal-raising-a-segmentation-fault-when-using-a-node-js-themin] suggested to remove the .info files.
I implemented this and it fixed all problems.


## Follow up
We need to merge the develop branch into all the broken build branches and let them run again on Travis-Ci.


## Extra
I also upgraded Drupal core & and contrib modules in the make files.

PR #751, opened 2015-05-21
CFD-611 Move to the latest version of OG module.

HelenaEksler:
https://issuetracker.amplexor.com/jira/browse/CFD-611


zero2one:
@HelenaEksler Needs a re-roll to fix the broken build

PR #752, opened 2015-05-21
CFD-571 search select scope

MPParsley:
- select the scope in which you want to search
- autocomplete search keywords

search


zero2one:
@MPParsley needs a re-roll

zero2one:
Entering a search term and hitting Enter does not start the search, you need to hit Enter twice
> This should be once.

Entering a search term and hitting the Search button does not start the search, you need to hit the button twice
> This should be once.

MPParsley:
@zero2one, this is the behaviour provided by Bootstrap:
/**
 * Prevents the form from submitting if the suggestions popup is open
 * and closes the suggestions popup when doing so.
 */
Drupal.autocompleteSubmit = function () {
  return $('.form-autocomplete > .dropdown').each(function () {
    this.owner.hidePopup();
  }).length == 0;
};

PR #763, opened 2015-05-27
CFD-526 wiki rearrange book pages

SebastienVanRintel:
image


zero2one:
@SebastienVanRintel : group administrators (and owners) should be able to rearrange pageswithin their group. Right now only site admins can rearrange.

Can you set the proper rights?

c4m-rearrange-wiki-permissions

PR #774, opened 2015-06-11
CFD-931-Admin_users_overview

SebastienVanRintel:
…tion view
image


zero2one:
Update table:

* Added first + last name.
* Grouped status fields.
* Grouped account dates.
* Added more filters.
* Added more sorting options.

c4m-user-admin

PR #783, opened 2015-06-18
CFD-893-Social_share

SebastienVanRintel:


SebastienVanRintel:
image

zero2one:
@SebastienVanRintel Add the social sharing buttons also to the Articles.

Same location as on the group content types.

PR #916, opened 2016-06-02
CFM-506 sharing between groups

soniCaH:


soniCaH:
https://issuetracker.amplexor.com/jira/browse/CFM-1237 needs to implement triggering a notification once this system is in place. A new PR will be created once this is finished.

PR #923, opened 2016-06-06
Create general unsubscribe functionality for notifications

ordavidil:
https://issuetracker.amplexor.com/jira/browse/CFM-1187


ordavidil:
Still need to fix the issue with the widget, @amitaibu any suggestions about this error:
selection_199

?

amitaibu:
Is that reference field indeed a regular reference?
What is the selection handler setting?

amitaibu:
Oh I see the selection handler is indeed base. You'll need to catch the exception and see why it thinks it is an og field. It might be related to a recent pach we have committed to OG

ordavidil:
`'base'` as you can see in here https://github.com/capacity4dev/capacity4more/pull/923/commits/16311d7f5cbc8571f2cd250a5c06af7ab1077355#diff-bdc33a563e8be04f4e2fd5326044f46aR529

ordavidil:
Basically [this one](https://github.com/capacity4dev/capacity4more/pull/923/commits/730527b9b0e519ed91bf53ac61a598762537b14a) is fixing it, but then the feature is overriden.

amitaibu:
actually, I don't think not related to recent patch. The exception is thrown from `og_field_widget_form` -- why/ where is that invoked is probably the first place to look.

ordavidil:
like, what I found that it is relate to the definition of the widget.

amitaibu:
* Don't use og widget - use a regular autocomplete
* To hide the field use `hook_field_access`

ordavidil:
:+1:

ordavidil:
@nader77 
can you review please ?
thanks

project/profiles/capacity4more/modules/c4m/user/c4m_user_profile/c4m_user_profile.module
```diff
@@ -844,3 +853,172 @@ function c4m_user_profile_init() {
...
+  if (!$account || c4m_user_is_spammer($account) || $account->status === 0) {
+    return FALSE;
+  }
+
+  $wrapper = entity_metadata_wrapper('user', $account);
+  if ($wrapper->c4m_unsubscribe_hash->value() != $hash) {
+    return FALSE;
+  }
+
+  if (!isset($wrapper->{"c4m_optout_{$entity_type}"})) {
```
nader77: Maybe add a small description of what is happening here.

nader77:
@ordavidil Looks good to me :)

project/profiles/capacity4more/modules/c4m/user/c4m_user_profile/c4m_user_profile.module
```diff
@@ -844,3 +853,173 @@ function c4m_user_profile_init() {
...
+  }
+
+  // Validate we have an optout field for the current entity type.
+  if (!isset($wrapper->{"c4m_optout_{$entity_type}"})) {
+    return FALSE;
+  }
+
+  // Validate entity ID.
+  try {
+    $wrapper = entity_metadata_wrapper($entity_type, $entity_id);
```
anvmn: @ordavidil 
I think there's an issue here for c4m_follow_group	 flag, as it follows a node.
entity_metadata_wrapper('group', [node_id]) will throw an exception, even if the values will be valid.
Also, c4m_follow_comment flag may be problematic as well, if Kevin confirms it should follow a node.

ordavidil: @anvmn 
we don't have an entity with the type `group`

anvmn: @ordavidil 
Right. But we have c4m_follow_group flag, and $entity_type is defined as flag suffix, to unsubscribe from.
Hence, you get entity_metadata_wrapper('group', [node_id]).

ordavidil: got it, good catch...

feel free to add this:
```php
// Since group is being handled by it's own flag we pass this value as the entity_type, hence for..... (keep going)
if ($entity_type == 'group') {
  $entity_type == 'node');
}
```
and add other logic that is necessary ...

project/profiles/capacity4more/modules/c4m/user/c4m_user_profile/c4m_user_profile.module
```diff
@@ -844,3 +853,173 @@ function c4m_user_profile_init() {
...
+  $wrapper->{$field}[] = $entity_id;
+  $wrapper->save();
+}
+
+/**
+ * Implements hook_message_subscribe_get_subscribers_alter().
+ *
+ * Remove unsubscribed users from the list when sending a notification.
+ */
+function c4m_user_profile_message_subscribe_get_subscribers_alter(&$uids, $values) {
```
anvmn: @ordavidil 
Is this hook activated when we provide $uids  $subscribe_options to message message_subscribe_send_message()? Code sample:
  $subscribe_options = array(
    'uids' => $flagging_uids, // array of users to be notified
  );
message_subscribe_send_message('node', $node, $wrapper->value(), array(), $subscribe_options);

ordavidil: Search for `drupal_alter('message_subscribe_get_subscribers', $uids, $values);` on the _message_subscribe.module_ file (should be line 248).

anvmn: I see... 
Module has to implement message_subscribe_get_subscriber(), and not pass $uids at  $subscribe_options for this hook to work.
Thanks,

PR #926, opened 2016-06-06
Notify user on membership activation with Email rework

anvmn:
No need to wrap the body with template - done by mimemail module.
selection_055


anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1414

PR #948, opened 2016-06-20
CFM-503 Pending projects

jover:


project/profiles/capacity4more/modules/c4m/admin/c4m_admin_projects/c4m_admin_projects.module
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Implements hook_menu().
+ */
+function c4m_admin_projects_menu() {
+  $items = array();
+
+  $items['admin/c4m/projects'] = array(
+    'title' => 'Projects',
+    'description' => 'Administration interface to manage projects & their status.',
+    'page callback' => 'views_embed_view',
```
amitaibu: wouldn't it be better to use View's own `page` display instead?

mfernea: I guess the purpose of this is to create a menu item and a subitem. admin/c4m/projects vs admin/c4m/projects/pending

PR #963, opened 2016-06-27
CFM-1446: Remove Media button from CKEditor for anonymous users.

mfernea:


soniCaH:
Added some margin between the disclaimer text and the captcha

PR #995, opened 2016-07-05
CFM-1444 - Fix sidebar js

soniCaH:
Merge #969 in develop (other way around causes weird issues)


soniCaH:
Changes fixed in original #969

PR #1030, opened 2016-07-17
[CFM-1067] Event Teaser View theming

guytepper:
https://issuetracker.amplexor.com/jira/browse/CFM-1067

Wireframe:
image

### DIFF
--- a/project/profiles/capacity4more/themes/c4m/kapablo/build/src/stylesheets/components/_events.scss
+++ b/project/profiles/capacity4more/themes/c4m/kapablo/build/src/stylesheets/components/_events.scss
@@ -53,17 +53,6 @@
       width: $icon-size;
     }
   }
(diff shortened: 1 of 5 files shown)


guytepper:
**Before:**


**After:**
screen shot 2016-07-17 at 1 55 27 pm

project/profiles/capacity4more/themes/c4m/kapablo/build/src/stylesheets/components/_events.scss
```diff
@@ -59,3 +59,12 @@
   margin-bottom: 10px;
 }
 
+.node-event.view-mode-teaser {
```
ordavidil: @guytepper 
I think this one can be nested under the previous `.node-event` class.

PR #1029, opened 2016-07-17
[CFM-907] Altering the body of the mail with the context of the group.

ordavidil:
https://issuetracker.amplexor.com/jira/browse/CFM-907

PR #1033, opened 2016-07-17
[CFM-908] Store gid+uid of a user after registration.

ordavidil:
https://issuetracker.amplexor.com/jira/browse/CFM-908


ordavidil:
Also destroyed the variables from the session.

ordavidil:
@janlemmens 
This one is dependent on #1029

PR #1035, opened 2016-07-18
[CFM-911] Associate user to group after login/activation.

ordavidil:
https://issuetracker.amplexor.com/jira/browse/CFM-911

PR #1044, opened 2016-07-19
[CFM-909] Redirect user to the group after activation.

ordavidil:
https://issuetracker.amplexor.com/jira/browse/CFM-909


ordavidil:
@janlemmens 
This one is dependent on #1035

PR #1045, opened 2016-07-19
[CFM-476] Send a notification when a group gets published

guytepper:
https://issuetracker.amplexor.com/jira/browse/CFM-476


guytepper:
Closed in favor of #1071 .

PR #1053, opened 2016-07-20
CFM-1205: Replaces custom alter hook with a hook that uses render arrays

DavidHernandez:
Each implementation on the hook can provide complex themes and provide a weight for the items added to the CTA block.

Jira issue: https://issuetracker.amplexor.com/jira/browse/CFM-1205

### DIFF
--- a/project/profiles/capacity4more/modules/c4m/user/c4m_user_og/c4m_user_og.module
+++ b/project/profiles/capacity4more/modules/c4m/user/c4m_user_og/c4m_user_og.module
@@ -462,10 +462,6 @@ function c4m_user_og_theme() {
  * Set the CTA block for visitors.
  */
 function c4m_user_og_c4m_cta_block() {
-  if (!user_is_anonymous()) {
-    return;
-  }
-
   if (!$group = c4m_og_current_group()) {
     return;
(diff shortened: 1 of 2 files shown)


project/profiles/capacity4more/modules/c4m/user/c4m_user_og/c4m_user_og.module
```diff
@@ -457,11 +457,15 @@ function c4m_user_og_theme() {
...
 
 /**
- * Implements hook_c4m_cta_block_alter().
+ * Implements hook_c4m_cta_block().
  *
  * Set the CTA block for visitors.
  */
-function c4m_user_og_c4m_cta_block_alter(&$block) {
+function c4m_user_og_c4m_cta_block() {
+  if (!user_is_anonymous()) {
```
ordavidil: @DavidHernandez 
this will cause a contradiction with:
```php
   if (!user_is_anonymous()) {
     global $user;
 
     if (og_get_membership('node', $group->nid, 'user', $user->uid)) {
       // Current user is already a member of this group.
       return;
     }
   }
```

which is down below (line 478).

PR #1048, opened 2016-07-19
WIP: [CFM-454] notify group admins about new membership request

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-454

PR #1054, opened 2016-07-20
Solved: Debugging behat test which fail on Travis

anvmn:


anvmn:
Notification was triggered during dummy content migration process, and could not find message_notify_send_message() function, because it's located at message_notify model, which is enabled after migration.

Fixed at: https://github.com/capacity4dev/capacity4more/pull/1048

PR #946, opened 2016-06-19
[CFM-1211] Set the uploaded photos filenames by the album.

ybaras:

PR #1070, opened 2016-07-24
[CFM-1211] Set the uploaded photos filenames by the album.

ordavidil:
https://issuetracker.amplexor.com/jira/browse/CFM-1211


ordavidil:
@soniCaH 
it seems that #946 was not merged to `develop` since it was compared to `feature/CFM-1211-photo-type-modifications` branch which probably has been merged before, hence, the code never made it to the `develop`.

This PR is here to fix that.

PR #1109, opened 2016-08-01
CFM-969: Open following for all users

DavidHernandez:
Allows authenticated users to follow/unfollow groups and projects

Issue: https://issuetracker.amplexor.com/jira/browse/CFM-969


jover:
How to test this? I see only permissions in the change logs...

ordavidil:
@DavidHernandez 
is there any change on the display for that ?
or maybe configuration that we can see at the admin ?

DavidHernandez:
@ordavidil @jover the issue only talks about the "follow" action to be available for all users. There is a different PR / Issue related with this one where the link is actually added and there can be tested that the link works for all the users: https://github.com/capacity4dev/capacity4more/pull/1110

So yes, maybe we have to merge the other PR first in order to be able to test this.

ordavidil:
@DavidHernandez 
it seems like #1110 should've compared with the current branch of this PR (like a dependency)

PR #1116, opened 2016-08-02
CFM-1192: Create view block to display suggested groups

DavidHernandez:
Updates the suggested groups block to display 5 items, order them alphabetically and send the user interests as default filters on the group page.

Issue: https://issuetracker.amplexor.com/jira/browse/CFM-1192

suggested-groups

### DIFF
--- a/project/profiles/capacity4more/modules/c4m/features/c4m_features_homepage/c4m_features_homepage.module
+++ b/project/profiles/capacity4more/modules/c4m/features/c4m_features_homepage/c4m_features_homepage.module
@@ -96,8 +96,8 @@ function c4m_features_homepage_c4m_suggested_groups_get_groups() {
  * Callback function to be used within uasort to order alphabetically nodes.
  */
 function c4m_features_homepage_node_sort_by_title($first_node, $second_node) {
-  // strcasecmp can't be used directly on uasort, as we need to get the node
-  // title.
+  // Function strcasecmp can't be used directly on uasort, as we need to get
+  // the node title.
   return strcasecmp($first_node->title, $second_node->title);
 }
 
--- a/project/profiles/capacity4more/modules/c4m/features/c4m_features_homepage/c4m_features_homepage.module
+++ b/project/profiles/capacity4more/modules/c4m/features/c4m_features_homepage/c4m_features_homepage.module
@@ -54,27 +54,51 @@ function c4m_features_homepage_c4m_suggested_groups_get_groups() {
       ->propertyCondition('nid', $user_groups['node'], 'NOT IN');
   }
 
+  $link_filters = array();
   if (!empty($interests)) {
     $query->fieldCondition('c4m_vocab_topic', 'tid', $interests, 'IN');
+
+    // Adds each interest to the exposed filters that will be pass to the
+    // groups page.
+    foreach ($interests as $key => $interest) {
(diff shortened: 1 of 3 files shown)


project/profiles/capacity4more/modules/c4m/features/c4m_features_homepage/c4m_features_homepage.module
```diff
@@ -54,30 +54,54 @@ function c4m_features_homepage_c4m_suggested_groups_get_groups() {
...
+    // groups page.
+    foreach ($interests as $key => $interest) {
+      $link_filters['f[' . $key . ']'] = 'c4m_vocab_topic:' . $interest;
+    }
   }
 
   $result = $query
     ->addTag('random')
-    ->range(0, 3)
+    ->range(0, 5)
```
ordavidil: @DavidHernandez 
why `5` ?
can we use constant ?

It seems that somewhere else we need 3 ...

project/profiles/capacity4more/modules/c4m/features/c4m_features_homepage/c4m_features_homepage.module
```diff
@@ -54,30 +54,54 @@ function c4m_features_homepage_c4m_suggested_groups_get_groups() {
...
     return;
   }
 
+  // After selecting 5 random groups, they are ordered alphabetically.
+  $nodes = array();
   foreach ($result['node'] as $group) {
-    $node_view = node_view(node_load($group->nid), 'block_list');
+    $nodes[] = node_load($group->nid);
+  }
+  uasort($nodes, 'c4m_features_homepage_node_sort_by_title');
```
ordavidil: @DavidHernandez 

isn't it better to sort it on the query itself ?

DavidHernandez: The query is already sorted as random. The MySQL sort acts before the query limit. In our case we need to sort first as random, limit the amount of items, sort the selected items alphabetically. This is not possible to do it in MySQL without a subquery or with two selects. In this case, I prefer to do the sort on the PHP side.

On the other hand, we could drop the random sort, but then, we will always return the same items.

project/profiles/capacity4more/modules/c4m/features/c4m_features_homepage/c4m_features_homepage.module
DavidHernandez: As far as I know, this pane is used only on the homepage. I could only find one occurrence in the code.

project/profiles/capacity4more/modules/c4m/features/c4m_features_homepage/c4m_features_homepage.module
```diff
@@ -54,30 +54,54 @@ function c4m_features_homepage_c4m_suggested_groups_get_groups() {
...
 
+  $link_filters = array();
   if (!empty($interests)) {
     $query->fieldCondition('c4m_vocab_topic', 'tid', $interests, 'IN');
+
+    // Adds each interest to the exposed filters that will be pass to the
+    // groups page.
+    foreach ($interests as $key => $interest) {
+      $link_filters['f[' . $key . ']'] = 'c4m_vocab_topic:' . $interest;
+    }
```
soniCaH: Isn't it easier to link to `/groups?filter=interests`?

PR #1094, opened 2016-07-28
CFM-791: Remove all traces

DavidHernandez:
Removes all the user cookies and all the information from the session when a user removes the account.

Issue: https://issuetracker.amplexor.com/jira/browse/CFM-791

PR #1122, opened 2016-08-02
CFM-791: Remove all traces (vs develop)

jover:
Original PR was #1094

Removes all the user cookies and all the information from the session when a user removes the account.

Issue: https://issuetracker.amplexor.com/jira/browse/CFM-791


DavidHernandez:
Tests are green again.

jover:
@DavidHernandez The "setcookie"-function doesn't seem to have effect on my local environment.
Is this possible or how did you test this?

DavidHernandez:
@jover I've been testing this deeply and I found a couple things that can affect this:
- Firefox ignores the cookie deletion or still displays expired cookies. In Chrome works fine.
- It's possible than the time of the client doesn't correspond with the time of the server. So, this could cause that the expiration time in the cookie set in the server has not arrived to the time in the client. I've solved this changing to an expiration date that is in the past for sure (-1).
- Drupal creates a new session cookie in every page request if there is no session cookie. It also creates another cookie call has_js. We are removing them but they are being created again (but for a different user, the anonymous user).

(I will leave the PR assigned to me until the CI turns green).

PR #1176, opened 2016-08-15
[CFM-932] Removing the events and documents from the quickpost.

ordavidil:
https://issuetracker.amplexor.com/jira/browse/CFM-932


ordavidil:
@nader77 
can you review this one please ?
thanks ...

nader77:
@ordavidil Please add screenshots of the updated quick-post :)

nader77:
@ordavidil I think there's no point of having the discussion button if it's the only option, what do you think ?

ordavidil:
@nader77 
True, but i think that since the logic is already implemented then it's better not to touch for now, we can handle it under https://issuetracker.amplexor.com/jira/browse/CFM-933

ordavidil:
selection_256

PR #1186, opened 2016-08-17
Feature/cfm 373 pending groups

mfernea:
Try again.


soniCaH:
#970 was merged in GH, but apparently was not successful, so re-did it here.

PR #1209, opened 2016-08-22
[CFM-1274] -search form is no longer in the group menu.

ybaras:
https://issuetracker.amplexor.com/jira/browse/CFM-1274


ybaras:
sig

PR #1210, opened 2016-08-22
[CFM-1036] Fix Activation mail, per MT18 requirements

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1036


anvmn:
selection_046

PR #1220, opened 2016-08-23
[CFM-939] Hide quick post for non-group visitors

guytepper:
https://issuetracker.amplexor.com/jira/browse/CFM-939

### DIFF
--- a/project/profiles/capacity4more/modules/c4m/restful/c4m_restful_quick_post/templates/quick-post-form.tpl.php
+++ b/project/profiles/capacity4more/modules/c4m/restful/c4m_restful_quick_post/templates/quick-post-form.tpl.php
@@ -5,12 +5,11 @@
  * Template to render the Quick post forms.
  */
 
+global $user;
+$og_context = og_context();
 
-  global $user;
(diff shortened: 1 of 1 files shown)


project/profiles/capacity4more/modules/c4m/restful/c4m_restful_quick_post/templates/quick-post-form.tpl.php
```diff
@@ -4,6 +4,13 @@
  * @file
  * Template to render the Quick post forms.
  */
+
+global $user;
```
ordavidil: @guytepper 
this kind of logic should never be in a template.
please move it to the preprocess function.

guytepper:
Seems like it's already implemented on [c4m_restful_quick_post.module](https://github.com/capacity4dev/capacity4more/blob/develop/project/profiles/capacity4more/modules/c4m/restful/c4m_restful_quick_post/c4m_restful_quick_post.module#L240-L247)

ordavidil:
so just check this is really working, if so then close the PR and comment on JIRA

PR #1246, opened 2016-08-26
CFM-1435: Create admin backend to define template project.

mfernea:


soniCaH:
Split up in https://github.com/capacity4dev/capacity4more/pull/1254 and https://github.com/capacity4dev/capacity4more/pull/1255 to fix travis errors.

PR #1267, opened 2016-08-31
[CFM-1318] Adds missing fields to the Topic content type

DavidHernandez:
Adds back missing fields to the topic content type.

Issue: https://issuetracker.amplexor.com/jira/browse/CFM-1318

### Added missing fields
add-topics

### Site Admin (mariecurie) can promote to frontpage
site-admin


ordavidil:
@DavidHernandez 
If I remember correct, reverting the feature after installation fix that, so, probably a problem with the feature.

DavidHernandez:
@ordavidil it was not only an issue resolved by reverting the feature. There was another issue of a field that changed the name and was not updated on the feature: https://github.com/capacity4dev/capacity4more/pull/1267/files#diff-5d3e539ed3fdec2b40b851d808ad6c73L393

ordavidil:
@DavidHernandez 
this field was added because that is the definition on the wiki:

https://wiki.amplexor.com/confluence/display/C4DNG/Content%3A+Topic

selection_282


The only thing that is not by the wiki is the name of the field, but this is because there is already a field with the name _c4m_related_articles_ but its cardinality is different, hence a new field was created in the spirit of the other two:
https://github.com/capacity4dev/capacity4more/blob/cfa82ff61e77bb2e0b5c0e0fd8520486bc4b2a79/project/profiles/capacity4more/modules/c4m/content/c4m_content_topic/c4m_content_topic.info#L57-L58

ordavidil:
@DavidHernandez 
this PR is basically covers CFM-1317

DavidHernandez:
What is left to do in this issue?

ordavidil:
@DavidHernandez ^^
https://github.com/capacity4dev/capacity4more/pull/1267#issuecomment-245251124

DavidHernandez:
@ordavidil then, not applying this PR will leave the original field (c4m_related_article_unlimited). So, should I do any other change?

ordavidil:
Closing in favor of https://github.com/capacity4dev/capacity4more/pull/1336

PR #1275, opened 2016-09-04
[CFM-191] -Unpublished pages do not appear under rearrange pages.

ybaras:
[CFM-191](https://issuetracker.amplexor.com/jira/browse/CFM-191)


ybaras:
cfm-191

PR #1295, opened 2016-09-08
[CFM-1229] Fixes issue that stops CKEditor to be loaded on a contact page

DavidHernandez:
Adds extra check on custom js to avoid crashing and stopping the rest of the Drupal behaviors to be executed.

Issue: https://issuetracker.amplexor.com/jira/browse/CFM-1229


project/profiles/capacity4more/behat/features/contact.feature
```diff
@@ -22,4 +22,4 @@ Feature: Contact
     Given I am logged in as user "alfrednobel"
     When I go to user "mariecurie" contact form
     Then I should see "Contact"
-    And I see the wysiwyg
+    And I should see the "#cke_edit-c4m-contact-user-body-value" element
```
DavidHernandez: @ordavidil, are we sure that the wysiwyg exists with this check?

The wysiwyg is added by javascript after the page is loaded. It could happen something before the wysiwyg is loaded preventing it to be added as it happened before.

ordavidil: @DavidHernandez 
The previous test passed even without the wysiwyg, hence I had to change it.

I've tested the current one 4 times while removing the wysiwyg and putting it back.

PR #1311, opened 2016-09-14
[BUGFIX] determine power user.

ordavidil:
https://issuetracker.amplexor.com/jira/browse/CFM-1257


ordavidil:
@soniCaH 
This PR should fix the problem from this morning.

PR #1319, opened 2016-09-15
[CFM-1535] do not notify event triggering user

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1535


ordavidil:
@anvmn 
it seems that this PR is taking care only about admins, plus i'm not sure this is the right approach...

anvmn:
@ordavidil 
Group of users is notified by message_subscribe_send_message(), which filters event user on it's own.

ordavidil:
@anvmn 
The documentation on this PR doesn't mention anything about not __notifying__ the event owner, only to filter, but we also provide the main reason for that filter.

project/profiles/capacity4more/modules/c4m/user/c4m_user_notifications/includes/c4m_user_notifications_handlers.inc
```diff
@@ -234,6 +243,8 @@ function _c4m_user_notifications_to_admin_group_status_set_republished($node, $a
  * Email notification about organic group status change to 'unpublished'.
  */
 function c4m_user_notifications_group_status_set_unpublished($node, $group_former_status) {
+  // Get the ids of group owner and all group admins. Filter out current user,
+  // since we don't want him to be notified on his own action.
   $group_admins_ids = c4m_og_get_group_admins_ids($node->nid, TRUE);
```
anvmn:
```php

/**
 * @param $nid
 * @return array
 */
function c4m_og_get_group_admins_ids_with_owner($nid) {
  return c4m_og_get_group_admins_ids($nid, TRUE);
}
```

PR #1328, opened 2016-09-18
Fixing new sniffer version error.

ordavidil:


ordavidil:
@soniCaH 
The coder module has been updated.
This one is fixing all the new errors we get from sniffer.

PR #1374, opened 2016-09-25
[CFM-943-fix] Flag content for any state, notify only if state is published

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-943


project/profiles/capacity4more/modules/c4m/user/c4m_user_notifications/c4m_user_notifications.module
```diff
@@ -250,12 +260,25 @@ function c4m_user_notifications_comment_insert($comment) {
...
+  }
+
+  $comment_notification_types = c4m_og_get_extended_content_notification_types();
   if (!$node->type || !in_array($node->type, $comment_notification_types)) {
     // Node, comment is associated with, does not require notification.
     return;
   }
 
+  $wrapper = entity_metadata_wrapper('node', $node);
+  if (!empty($wrapper->{OG_AUDIENCE_FIELD}) && $wrapper->{OG_AUDIENCE_FIELD}->c4m_og_status->value() != 'published') {
```
ordavidil: @soniCaH 
Please confirm this one since there is a big impact on all comments notifications

soniCaH: @ordavidil 
Do I interpret this right that this prevents notifications for comments (on content) made on draft/pending/archived/deleted groups?
If so, I think this is indeed right!

PR #1382, opened 2016-09-26
[CFM-1476] Broken document widget

ybaras:
[CFM-1476](https://issuetracker.amplexor.com/jira/browse/CFM-1476)

PR #1401, opened 2016-09-28
Feature/cfm 1476 document widget dnm1

ybaras:


ybaras:
No longer needed. See #1382 .

PR #1407, opened 2016-09-30
Redirect taxonomy term pageviews to the global search page + Create breadcrumbs for photos and photoalbums

soniCaH:


soniCaH:
Photoalbum:

screen shot 2016-09-30 at 15 25 31

Photo without album:

screen shot 2016-09-30 at 15 19 29

Photo with album:

screen shot 2016-09-30 at 15 19 25

PR #1433, opened 2016-10-18
[CFM-1516] Fix news in the global Activity Stream

ybaras:
[CFM-1516](https://issuetracker.amplexor.com/jira/browse/CFM-1516)


ybaras:
**Members can now see news items in the global activity stream.**

globalactivity

PR #1436, opened 2016-10-20
WIP: [CFM-1254-fix] Cancel notification, when content is highlighted.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1254


project/profiles/capacity4more/modules/c4m/content/c4m_content/c4m_content.module
```diff
@@ -786,6 +786,10 @@ function c4m_content_toggle_highlight_access($group, $node, $status) {
 function c4m_content_toggle_highlight($group, $node, $status, $ajax = 'nojs') {
   $wrapper = entity_metadata_wrapper('node', $node);
 
+  // Add parameter to node that will signal to cancel  notification that is
+  // triggered, as a result of node update.
```
ordavidil: ... because we should not send a notification for highlight ...  (or something like that)

PR #1432, opened 2016-10-18
[CFM-937-fix] remove fields from QuickPost form

anvmn:
## Merge #1443 First

https://issuetracker.amplexor.com/jira/browse/CFM-937

PR #1443, opened 2016-10-25
[CFM-940] Adjust QP form for accessibility.

anvmn:
## Branch from #1432

https://issuetracker.amplexor.com/jira/browse/CFM-940

### DIFF
--- a/project/profiles/capacity4more/themes/c4m/kapablo/build/src/stylesheets/components/_quickpost.scss
+++ b/project/profiles/capacity4more/themes/c4m/kapablo/build/src/stylesheets/components/_quickpost.scss
@@ -31,6 +31,10 @@
       right: 7px;
 
       label {
+        color: transparent;
+      }
+
+      i {
         color: $color-text-alternative;
         text-shadow: 1px 1px 1px $color-gray-light;
         font-size: 1.5em;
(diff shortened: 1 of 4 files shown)


anvmn:
selection_112

project/profiles/capacity4more/themes/c4m/kapablo/build/src/stylesheets/components/_quickpost.scss
```diff
@@ -31,6 +31,10 @@
       right: 7px;
 
       label {
+        color: transparent;
```
ordavidil: why do we need a transparent color for the label ?

anvmn: The label contains `<i>` tag that - attachment icon.
If we use `visibility: hidden`, it will be hidden.

`transparent color` is to hide label text, which is required for accessibility only.

anvmn: I tried this, but this way, when clicking on `<i>` icon, `<a>` link above the label is not activated:

   `<a href="javascript://" ng-click="browseFiles('c4m-related-document')"></a>
    <span class="body-attachment-link">
        <label for="c4m-related-document">
         <i class="fa fa-paperclip"></i>
       </label>`
    </span>
  </div>

PR #1467, opened 2016-11-01
[CFM-1192-fix] Don't suggest groups for user that got no interests.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1192

### DIFF
--- a/project/profiles/capacity4more/behat/features/homepage.feature
+++ b/project/profiles/capacity4more/behat/features/homepage.feature
@@ -64,8 +64,8 @@ Feature: Test homepage content and blocks
     Then  I should not see "Suggested Groups"
 
   @api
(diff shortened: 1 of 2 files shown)


project/profiles/capacity4more/behat/features/homepage.feature
```diff
@@ -64,8 +64,8 @@ Feature: Test homepage content and blocks
     Then  I should not see "Suggested Groups"
 
   @api
-  Scenario: Logged in, non member user should see "Suggested Groups" block
-    Given I am logged in as user "president"
```
ordavidil: @anvmn 
if you're already here then add a test with _president_ that verify we do not see the block

anvmn: @ordavidil  Done

PR #1474, opened 2016-11-02
[CFM-1220-fix] Do not notify users that are banned at group.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1220


ordavidil:
Closing in favor of #1475

PR #1471, opened 2016-11-02
[CFM-1229-fix] Fix redirect url and status message

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1229

PR #1490, opened 2016-11-07
[CFM-1229-fix] Fix redirect url and status message.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1229

## Replaces #1471


anvmn:
out

PR #1512, opened 2016-11-11
CFM-1236: Adjust displayed info for shared content.

mfernea:


project/profiles/capacity4more/modules/c4m/content/c4m_content_topic/c4m_content_topic.info
```diff
@@ -84,7 +83,6 @@ features[variable][] = node_preview_topic
 features[variable][] = node_submitted_topic
 features[variable][] = pathauto_node_topic_pattern
 features[views_view][] = c4m_expert_users
-features[views_view][] = c4m_message_with_related_topics_no_js_fallback
```
ordavidil: @mfernea 
why is this view is being deleted ?

soniCaH: The topic page is already displaying the activity stream as a regular view block. It's not using any of the restful components at all.
Basically we were rendering the exact same view twice, once with full pager (hidden on JS), once with ajax pager (hidden when no JS). The ajax pager however already has a no-JS fallback.

ordavidil: As far as we've checked it was not implemented, if it got fixed after then np.
just want to make sure that this is not done by mistake.

@ybaras 
please confirm

PR #1527, opened 2016-11-14
[CFM-1265-fix] Initial project state

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1265


project/profiles/capacity4more/modules/c4m/field/c4m_field_og/c4m_field_og.module
```diff
@@ -119,7 +119,10 @@ function c4m_field_og_field_access($op, $field, $entity_type, $entity, $account)
  * Hides certain status values at form, according to users access rights.
  */
 function c4m_field_og_form_group_node_form_alter(&$form, &$form_state, $form_id) {
-  _c4m_field_og_set_status_flow($form, $form_state, 'group');
+  // If editing node, set the status flow.
+  if (isset($form['nid']['#value'])) {
```
ordavidil: @anvmn 
what's the problem in here ? the comment is lacking ...

ps
not sure how's the group content type is related to this PR.

anvmn: Status flow sets allowed values for 'c4m_og_status' field, as described at following diagram.
The problem here is that the action is supposed to execute on edit, and the alter is triggered on create, as well.

Don't think that group content type is related here. Why do you ask about it?

ordavidil: still don't understand the change that was made ...

PR #1531, opened 2016-11-14
[CFM-1138] Remove title of fallback js activity stream.

ordavidil:
https://issuetracker.amplexor.com/jira/browse/CFM-1138

selection_323


ordavidil:
The failure test is problematic in general and is not related to this PR.

PR #1557, opened 2016-11-17
CFM-1566: Project administration and permissions.

soniCaH:
MERGE #1558 FIRST


project/profiles/capacity4more/modules/c4m/og/c4m_og_group_menu/c4m_og_group_menu.features.menu_links.inc
```diff
@@ -20,7 +20,7 @@ function c4m_og_group_menu_menu_default_menu_links() {
       'attributes' => array(
         'title' => 'Configuration for Organic groups menu',
       ),
-
+      'identifier' => 'management_og-menu-settings:admin/config/group/og_menu',
```
ordavidil: @soniCaH 
will this reflect on #1553

ordavidil:
@soniCaH 
It seems that there is a preference for a _member_ over _administer member_ for projects permissions.
Is that necessary ?
I hope it won't break anything ...

soniCaH:
@ordavidil 
Problem is that we don't have "regular" members in projects. If you are a member of a project, you should be Project Administrator.

I think we have two options:

1. Remove the administrator member role for projects and map all permissions on regular member (as I did in this PR)
2. Keep the same permissions in groups as in projects, but then we have to implement hooks to make sure if a user is added to a project, we automatically give him the "administrator member" role.

With the last option we can end up with situations that a project owner can remove this role, and we end up with "orphan" members in projects.

Do you think the second option will cause less potential issues?

ordavidil:
My concern is regarding places on the platform that check if a user is power user in order to take actions ...

ordavidil:
I guess we can alter that function to return `TRUE` if the group type is `project` and the user is a member.
But I hope there are no other places ...

soniCaH:
I know, had to check them one by one.

Was having a look at OG User Roles, where you can define a default user role, but seems like development stopped long long time ago.

Do you have any concerns (other than the possibility to remove the role) about the hook implementation?
```
hook_og_membership_insert() {
  if ($membership->...->type == 'project') {
    grant_role($user, 'administrator member');
  }
}
```

PR #1581, opened 2016-11-21
CFM-1264: Implement hook_node_access_records for project type as well.

soniCaH:


ordavidil:
@soniCaH 
I'm closing this one since #1612 handled it already ...

open if you think we still need it.

PR #1585, opened 2016-11-22
[CFM-1255-fix] Configure photo and photoalbum CTs for group dashboard highlight

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1255


anvmn:
selection_158

PR #1593, opened 2016-11-24
[CFM-1344-fix] Do not apply bold font on control-label at node form

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1344


anvmn:
selection_171

PR #1594, opened 2016-11-24
[CFM-1255-fix] photo and photoalbum CTs - conifugre block list views

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1255


anvmn:
selection_172

PR #1598, opened 2016-11-27
[CFM-486-fix] Verify 'email-change confirmation' mails format is defined per user profile.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-486

### DIFF
--- a/project/profiles/capacity4more/modules/c4m/user/c4m_user_profile/c4m_user_profile.module
+++ b/project/profiles/capacity4more/modules/c4m/user/c4m_user_profile/c4m_user_profile.module
@@ -304,6 +304,12 @@ function c4m_user_profile_user_login(&$edit, $user) {
  * sets 'email confirmation' mail format accordingly.
  */
 function c4m_user_profile_mail_alter(&$message) {
+  // Mime mail module converts notification mails to 'plain text' format,
+  // in case user enabled this option at his profile (loads user profile by
+  // mail, and checks if 'mimemail_textonly' is enabled).
+  // 'change mail' confirmation notification is special case, since it's not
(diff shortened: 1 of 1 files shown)


ordavidil:
@anvmn 
screenshot / gif

anvmn:
out

project/profiles/capacity4more/modules/c4m/user/c4m_user_profile/c4m_user_profile.module
```diff
@@ -298,6 +298,21 @@ function c4m_user_profile_user_login(&$edit, $user) {
 }
 
 /**
+ * Implements hook_mail_alter().
+ *
+ * For users that requested email notifications format to be plaintext,
+ * sets 'email confirmation' mail format accordingly.
+ */
+function c4m_user_profile_mail_alter(&$message) {
+  if ($message['id'] != 'email_confirm_confirmation_email') {
```
ordavidil: @anvmn 
this should work for all emails that the user receive from the system, not only `email conform confirmation email`

anvmn: Mime mail module takes care of it - `mimemail_prepare_message()` function.
It loads the $account (load by mail) and checks if it has palin text option enabled.
Confirmation of new mail is special case, because it's impossible to load user by new mail - it does not exist.
Therefore, I add special logic only for this case.

ordavidil: ok, so please add this as a comment so anyone can understand that ...

PR #1608, opened 2016-11-30
[CFM-1307] Add project highlight icon to project manage cta block.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1307


anvmn:
out

ordavidil:
@anvmn 
check the test with that scenario:
`Scenario Outline: SA highlight a group via its dashboard.`

and add a similar one for a project as well (you can make it a regular one and not outline - 1 SA should be enough).

PR #1613, opened 2016-12-04
WIP: Replace callback function for message params, with their resolved value.

anvmn:

PR #1614, opened 2016-12-04
Replace callback function for message params, with their resolved value.

anvmn:
Code from #1613

### DIFF
--- a/project/profiles/capacity4more/modules/c4m/message/c4m_message/c4m_message.module
+++ b/project/profiles/capacity4more/modules/c4m/message/c4m_message/c4m_message.module
@@ -548,6 +548,21 @@ function c4m_message_entity_load($entities, $entity_type) {
 }
 
 /**
+ * Replaces callback function for message params, with their resolved value.
+ *
+ * @param Message $message
+ *   The Message entity.
+ */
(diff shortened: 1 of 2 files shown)


ordavidil:
@anvmn 
it seems that not all messages are covered, search for `message_create(` and catch the rest of the scenarios as well ...

anvmn:
@ordavidil 
I found one place here: https://github.com/capacity4dev/capacity4more/pull/1614/commits/bf77dee650dc37623b8ec2bbddeacbceba3a3cd0#diff-a48c9f232aa5ef72e9f1d837774655daL1068

Did you see others?

ordavidil:
@anvmn - here's a simple diff:

selection_350

### VS

selection_351


so just make sure we cover it all ...

anvmn:
@ordavidil 
There are messages created for activity streams.They are not related to notification system.
Notifications message are concentrated at c4m_user_notifications_handlers.inc file.

As I see it, all messages are covered.

ordavidil:
The issue is for all created messages, not only notifications ...

soniCaH:
@ordavidil 

screen shot 2016-12-09 at 14 26 37

PR #1629, opened 2016-12-26
[CFM-1561] Fixing invite issues.

ordavidil:
https://issuetracker.amplexor.com/jira/browse/CFM-1561

- Fixed the decline message by adding a patch.
- Fixed the title and the button on the registration form.
- Fixed a bug when a user try to invite a member to a specific group but withing a context of a different group (can happen if changing the url explicitly).
- Removed the _always decline_ option since it has a [bug](https://www.drupal.org/node/2320373) on the contrib itself.

### DIFF
--- a/config/make/og.make
+++ b/config/make/og.make
@@ -37,7 +37,7 @@ projects[session_api][version] = "1.0-rc1"
 
 projects[og_invite][subdir] = "contrib"
 projects[og_invite][version] = "1.0-beta5"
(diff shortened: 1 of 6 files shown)


ordavidil:
@soniCaH 
Any idea why travis keep fails ?
Is there something that has been changed ?

selection_354

ordavidil:
Sorry, it way my bad with wrong path on the patch ....

PR #1640, opened 2017-01-22
[CFM-966-fix] Do not create AS for Draft content.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-966

### DIFF
--- a/project/profiles/capacity4more/modules/c4m/message/c4m_message/c4m_message.module
+++ b/project/profiles/capacity4more/modules/c4m/message/c4m_message/c4m_message.module
@@ -3230,16 +3230,6 @@ function c4m_message__insert__node($node) {
  *   The non-saved message entity, or NULL if update shouldn't occur.
  */
 function c4m_message__update__node($node, $account = NULL) {
-  if (empty($account)) {
-    global $user;
-    $account = clone $user;
-  }
-
-  if ($account->uid == 0) {
-    // Drush has user "0". Assign updates to the original node creator.
-    $account = user_load($node->uid);
-  }
-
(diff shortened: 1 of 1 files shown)


project/profiles/capacity4more/modules/c4m/message/c4m_message/c4m_message.module
```diff
@@ -3204,6 +3204,12 @@ function c4m_message__update__comment($comment) {
  *   The non-saved message entity.
  */
 function c4m_message__insert__node($node) {
+  // Do not create message for activity stream, as we deal with
+  // unpublished content.
+  if ($node->status == NODE_NOT_PUBLISHED) {
```
ordavidil: should we use `===` ?

project/profiles/capacity4more/modules/c4m/message/c4m_message/c4m_message.module
```diff
@@ -3234,9 +3240,17 @@ function c4m_message__update__node($node, $account = NULL) {
     $account = user_load($node->uid);
   }
 
+  // Do not create/update message for activity stream, as we deal with
```
ordavidil: move this to be the first condition to check, there is no point to load a user if we return early ...

project/profiles/capacity4more/modules/c4m/message/c4m_message/c4m_message.module
anvmn: `==` seems to be sufficient for node status check (went through code that checks for node status, and it's always `==`).

PR #1642, opened 2017-01-25
[CFM-492-fix] Make sure toggle link includes purl, so we get the OG context.

anvmn:
Another fix for https://issuetracker.amplexor.com/jira/browse/CFM-492.


project/profiles/capacity4more/modules/c4m/content/c4m_content/c4m_content.module
```diff
@@ -1194,7 +1194,7 @@ function c4m_content_create_toggle_group_notification_link($account, $group) {
     'attributes' => array(
       'class' => 'use-ajax',
     ),
-    'purl' => array('disabled' => TRUE),
+    'purl' => array('disabled' => FALSE),
```
ordavidil: you may want to use `'enabled => TRUE`

PR #1658, opened 2017-02-06
[CFM-630-fix] Sort vocab widget terms by 'weight'.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-630
https://issuetracker.amplexor.com/jira/browse/CFM-1624

### DIFF
--- a/project/profiles/capacity4more/modules/c4m/vocab/c4m_vocab_widget/c4m_vocab_widget.module
+++ b/project/profiles/capacity4more/modules/c4m/vocab/c4m_vocab_widget/c4m_vocab_widget.module
@@ -193,6 +193,7 @@ function c4m_vocab_widget_prepare_taxonomy_tree(array $terms) {
         // This is the parent term.
         $options[$term->tid]['label'] = htmlspecialchars_decode(filter_xss($term->name));
         $options[$term->tid]['id'] = $term->tid;
+        $options[$term->tid]['weight'] = $term->weight;
       }
       else {
         $id = $parents[0];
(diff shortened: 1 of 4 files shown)


anvmn:
anim

soniCaH:
@anvmn quick note: they should not be sorted alphabetically, they should be sorted by the weight they have in the drupal backend.

anvmn:
@soniCaH 
As I see it, it will be much harder to find desired value, If terms are not sorted alphabetically.
What's the logic behind the weight?

soniCaH:
Don't know I understand the question 100%.
It's the default Drupal weight property on the $taxonomy objects, so it goes from a negative value (highest) to the largest positive value (displayed lowest)

ordavidil:
@soniCaH 
taking the discussion back to JIRA

project/profiles/capacity4more/modules/c4m/vocab/c4m_vocab_widget/c4m_vocab_widget.module
```diff
@@ -193,7 +193,7 @@ function c4m_vocab_widget_prepare_taxonomy_tree(array $terms) {
         // This is the parent term.
         $options[$term->tid]['label'] = htmlspecialchars_decode(filter_xss($term->name));
         $options[$term->tid]['id'] = $term->tid;
-        $options[$term->tid]['weight'] = $term->weight;
+        $options[$term->tid]['weight'] = (int)$term->weight;
```
ordavidil: add a space between the cast operator and the value.

PR #1665, opened 2017-02-14
[CFM-422-fix] Set node_insert message format to 'mail'.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-422


anvmn:
anim

PR #1668, opened 2017-02-14
[CFM-624-fix] Redirect path when updating group purl.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-624


anvmn:
selection_446

PR #1671, opened 2017-02-15
[CFM-535-fix] Show 'add wiki' button, as if there're no wiki pages, even if unpublished wiki pages exist.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-535


anvmn:
Tennis group got a draft wiki page.

selection_449

PR #1677, opened 2017-02-22
[CFM-374-fix] Do not update group content realms when group request is approved/rejected.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-374

### DIFF
--- a/project/profiles/capacity4more/modules/c4m/og/c4m_og/c4m_og.module
+++ b/project/profiles/capacity4more/modules/c4m/og/c4m_og/c4m_og.module
@@ -1890,9 +1890,11 @@ function c4m_og_handle_group_status_change($node) {
   $status = $wrapper->c4m_og_status->value();
   $former_status = $wrapper_original->c4m_og_status->value();
 
-  if ($status != $former_status) {
-    // Use OG Access mechanism to update realms of group content.
-    // Update will be performed using Batch API.
+  // On status change, use OG Access mechanism to update realms of group
+  // content. Update will be performed using Batch API.
+  // Not required when group is approved (status changes from 'pending', as
+  // group got no content at that point.
+  if ($status != $former_status && $former_status != 'pending') {
     og_access_handle_group_privacy_change('node', $node);
   }
 }


anvmn:
anim

ordavidil:
@anvmn 
will we have access issues after this?
please verify with all roles (SA/GO/ etc.)

anvmn:
@ordavidil 
There  `c4m_og_handle_group_status_change()` was added at https://github.com/capacity4dev/capacity4more/pull/1634.
It assigns group realms to group content, when group status changes.
This was done to present correct content at views, if  I remember correctly.
As 'pending' group got no content, there's no need to execute the function.

Don't see access issue here.

soniCaH:
@anvmn do I understand this remark correctly?

> As 'pending' group got no content, there's no need to execute the function.

It is possible to add content to a pending group, to "prepare" it before being approved. So by the time a group is approved, it is possible it has already plenty of content.

ordavidil:
@soniCaH 
As I remember from talks with Diane, content can not be added to a pending group, but only to a draft one.

soniCaH:
@ordavidil the draft status was replaced by pending when we started this phase.
When you request a new group, you are able to add content.

ordavidil:
@soniCaH not sure this is right, as I remember (and this one pretty good), _pending_ replaced the _requested_ status.

soniCaH:
@ordavidil seems right. Must have tested it with mariecurie which is a SA then.
Bless your memory.

The only potential issue is a SA adding content to a pending group before approving it then?

ordavidil:
Which is very strange situation that should not happen.

ordavidil:
I think we also discussed about this with Diane if I remember correctly - specifically about SA adding content to pending group,  @anvmn do you remember this conversation >

anvmn:
@soniCaH 
We had a discussion about this with Diane, and final decision was that no one is allowed to add content to pending group. Not even site admin.
Final permissions summary:
selection_476

soniCaH:
@anvmn I see, the issue mainly is that the "add document", "add event" buttons + "quickpost"... are all shown in a pending group.

soniCaH:
screen shot 2017-02-22 at 18 13 59

ordavidil:
@soniCaH 
The buttons will lead to access denied - which is fine.

@anvmn 
Check the quickpost again, if needed then make the changes in the `checkEntityAcess()` under the `C4mRestfulEntityBaseNode` class (_c4m_restful_ module).

anvmn:
@ordavidil 
I see that admin can create a discussion via quick post.
I think that if such action is not allowed, QP form should not be visible at all.
If I'm not mistaken, that is what we do when user may not post with QP create group content.

ordavidil:
@soniCaH 
It seems like a sufficient solution for me,  please confirm.

ordavidil:
@soniCaH ^^

soniCaH:
@ordavidil @anvmn agreed.

ordavidil:
@anvmn 
please open a new PR for that.

PR #1680, opened 2017-02-23
[CFM-374-fix] Hide QP at 'pending' group, if user is site admin.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-374

### DIFF
--- a/project/profiles/capacity4more/modules/c4m/restful/c4m_restful_quick_post/c4m_restful_quick_post.module
+++ b/project/profiles/capacity4more/modules/c4m/restful/c4m_restful_quick_post/c4m_restful_quick_post.module
@@ -272,8 +272,16 @@ function c4m_restful_quick_post_form($group_id) {
     }
   }
 
+  // Site admin may not create content at pending and deleted groups.
+  // Special check required as node_access() allows everything to site admin.
   if (c4m_user_is_site_admin()) {
-    $user_may_create_discussion = c4m_og_node_access('discussion', 'create', $user) != 'deny';
+    $group = c4m_og_current_group();
+    $group_state = c4m_og_get_group_status($group);
+
+    $user_may_create_discussion = !in_array($group_state, array(
+      'pending',
+      'deleted',
+    ));
   }
   else {
     $user_may_create_discussion = node_access('create', 'discussion');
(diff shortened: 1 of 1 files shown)


project/profiles/capacity4more/modules/c4m/restful/c4m_restful_quick_post/c4m_restful_quick_post.module
```diff
@@ -272,9 +272,12 @@ function c4m_restful_quick_post_form($group_id) {
     }
   }
 
+  $user_may_create_discussion = c4m_user_is_site_admin() ?
```
ordavidil: @anvmn 
please replace this with a simple `if` condition.

ordavidil: also, the variable should probably relate to quickpost (so rename the variable)

project/profiles/capacity4more/modules/c4m/restful/c4m_restful_quick_post/c4m_restful_quick_post.module
```diff
@@ -272,9 +272,16 @@ function c4m_restful_quick_post_form($group_id) {
     }
   }
 
+  if (c4m_user_is_site_admin()) {
```
ordavidil: @anvmn 
please add an elaborated comment.

project/profiles/capacity4more/modules/c4m/restful/c4m_restful_quick_post/c4m_restful_quick_post.module
anvmn: This variable indicates if user has a permission to create discussion.
Comment that comes next explains how it's used with relation to QP form:

```
  // If all resources are turned off, or user is not allowed to create a
  // discussion, disable the QP form.
```

ordavidil: not sure which resources you're referring to, but, by elaborating I meant to explain __when__ the user is not allowed.
The fact that the user is not allowed is trivial from the code.

PR #1687, opened 2017-02-27
[CFM-1590-fix] Use custom files_upload RESTful plugin.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-1590


anvmn:
@ordavidil 
This PR uses custom plugin instead of generic, to avoid the warnings.

ordavidil:
Closing since this is not relevant.

PR #1705, opened 2017-03-05
[CFM-654-fix-2] News comment display layout.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-654


anvmn:
After clean installation:

selection_524

PR #1706, opened 2017-03-05
[CFM-403-fix] Share icons for private/restricted groups.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-403


anvmn:
**Private:**
selection_530

**Public:**
selection_531

**Restricted:**
selection_532

PR #1713, opened 2017-03-13
[CFM-292-fix] Set c4m_banner scheme as public.

anvmn:
https://issuetracker.amplexor.com/jira/browse/CFM-292


anvmn:
selection_558

PR #1766, opened 2017-12-08
CFDRMD-128: Refactor & extend notifications system

soniCaH:


soniCaH:
@mfernea the cronjob is handled by drupal queue. You should see a `queue_message_subscribe` entry.

PR #1787, opened 2018-01-02
CFDRMD-59: Notifications for topics/regions of interest

soniCaH:
Please review #1789 first!


project/profiles/capacity4more/modules/c4m/user/c4m_user_subscribe/c4m_user_subscribe.install
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * @file
+ * Install and update hooks for the c4m_user_subscribe functionality.
+ */
+
+/**
+ * Implements hook_update_N().
+ */
+function c4m_user_subscribe_update_7001(&$sandbox) {
+  $ret = [];
```
mfernea: This variable is no longer used.

project/profiles/capacity4more/modules/c4m/user/c4m_user_subscribe/c4m_user_subscribe.install
```diff
@@ -0,0 +1,90 @@
...
+        '@count_terms' => $sandbox['max'],
+      ]);
+  }
+}
+
+/**
+ * Implements hook_update_N().
+ */
+function c4m_user_subscribe_update_7002(&$sandbox) {
+  $ret = [];
```
mfernea: The content of this function is very similar to the one above. Only the vocab machine name and the name used in the message differs. Although this code is ran once and doesn't need maintenance, I would create a single function with params which is called twice.

project/profiles/capacity4more/modules/c4m/message/c4m_message/c4m_message.module
```diff
@@ -3350,6 +3403,36 @@ function c4m_message_format_node_share(Message $message) {
...
+ *
+ * @return string
+ *   The unsubscribe link.
+ */
+function c4m_message_format_unsubscribe_topic_url(Message $message) {
+  return _c4m_message_format_unsubscribe_url($message, 'taxonomy_term');
+}
+
+/**
+ * Return the formatted share operation.
```
mfernea: Is the comment correct? Doesn't seem so.

soniCaH: Fixed. At first we only needed notifications for topics. Generalised the naming now.

project/profiles/capacity4more/modules/c4m/user/c4m_user_notifications/includes/c4m_user_notifications.notify_subscribe.inc
```diff
@@ -218,6 +218,115 @@ function _c4m_user_notifications_notify_subscribe_group_content($node, $action,
...
+    // 'revision access' permission.
+    if (isset($node->vid) && isset($node->original->vid) && $node->vid !== $node->original->vid) {
+      $subscribe_options = _c4m_user_notifications_get_skip_entity_access_options();
+    }
+    // Message will be saved in message_subscribe_send_message().
+    message_subscribe_send_message('taxonomy_term', $term, $wrapper->value(), [], $subscribe_options, $context);
+  }
+
+  // Iterate over all tagged geo terms.
+  foreach ($node_wrapper->get('c4m_vocab_geo')->value() as $term) {
```
mfernea: For this loop and the above one I think we should use a single function with a param, looks like a lot of duplicated code now.

project/profiles/capacity4more/modules/c4m/user/c4m_user_notifications/includes/c4m_user_notifications.helper.inc
```diff
@@ -74,19 +74,49 @@ function _c4m_user_notifications_message_subscribe_get_subscribers(Message $mess
...
+    else {
+      $group = NULL;
+    }
+  }
 
   // Filtering users and preparing final user list.
   $users = user_load_multiple(array_keys($users));
+
+  // Already subscribed users.
+  $previous_recipients = _c4m_user_notifications_get_message_unique_recipients($message->mid);
```
mfernea: I think it's best to get and use $previous_recipients before user_load_multiple() call. So that we load less users.

PR #1819, opened 2018-02-02
CFDRMD-64: Create message digest with user interface.

soniCaH:


project/profiles/capacity4more/modules/c4m/message/c4m_message_digest/c4m_message_digest.install
```diff
@@ -0,0 +1,65 @@
...
+        'default' => '',
+      ],
+    ],
+    'indexes' => [
+      'fid' => ['fid'],
+      'entity_id' => ['entity_id'],
+      'uid' => ['uid'],
+    ],
+    'foreign keys' => [
+      'flag' => [
```
mfernea: Is this ok?

project/profiles/capacity4more/modules/c4m/message/c4m_message_digest/c4m_message_digest.install
```diff
@@ -0,0 +1,65 @@
...
+      ],
+      'preference' => [
+        'description' => 'The machine name of the digest preference.',
+        'type' => 'varchar',
+        'length' => 128,
+        'not null' => TRUE,
+        'default' => '',
+      ],
+    ],
+    'indexes' => [
```
mfernea: Shouldn't we add a primary key? Shouldn't we add the entity type as part of the key? Anyway is used in the search so I think the index will not be used.

project/profiles/capacity4more/modules/c4m/message/c4m_message_digest/includes/c4m_message_digest_preferences.inc
```diff
@@ -0,0 +1,137 @@
...
+   * Update the preference and return a new value object.
+   *
+   * @param string $preference
+   *   The machine name of the (new) digest preference.
+   *
+   * @return \C4mMessagePreference
+   *   New instance of the value object.
+   */
+  public function updatePreference($preference) {
+    return new C4mMessagePreference($this->fid, $this->entityType, $this->entityId, $this->uid, $preference);
```
mfernea: Why a new object is created?

project/profiles/capacity4more/modules/c4m/user/c4m_user_notifications/includes/c4m_user_notifications.helper.inc
```diff
@@ -106,9 +108,29 @@ function _c4m_user_notifications_message_subscribe_get_subscribers(Message $mess
...
+          $user_object = ['notifiers' => [$preference->getPreference() => $preference->getPreference()]];
+        }
+        else {
+          // No specific preference; use immediate notification.
+//          $user_object = ['notifiers' => ['email' => 'email']];
+
+          unset($users[$key]);
+          continue;
+        }
+      }
```
mfernea: I think it's best to add this to a separate function. The current function is already big aprox 100 lines.

project/profiles/capacity4more/modules/c4m/message/c4m_message_digest/includes/c4m_message_digest_preferences.inc
soniCaH: The object is immutable (value object). Putting in practice some best practices from DDD.

project/profiles/capacity4more/modules/c4m/message/c4m_message_digest/c4m_message_digest.install
soniCaH: @mfernea do you think there is an advantage in creating a dedicated key column? We will never load / reference a preference by this id.

project/profiles/capacity4more/modules/c4m/message/c4m_message_digest/c4m_message_digest.install
soniCaH: missed this when copy/pasting from another foreign key. Should be "flag" table of course.

project/profiles/capacity4more/modules/c4m/message/c4m_message_digest/c4m_message_digest.install
mfernea: We can add indexes, primary keys or unique keys. It's best to add a composed primary key rather than a composed index. It will ensure uniqueness, which is not enforced at the momment. I would also include the entity_type column in the key for the above reasons.

project/profiles/capacity4more/modules/c4m/user/c4m_user_profile/c4m_user_profile.pages.inc
```diff
@@ -149,3 +149,426 @@ function c4m_user_profile_remove_account_form_submit($form, &$form_state) {
...
+      'class' => [
+        'btn',
+        'btn-primary',
+      ],
+    ],
+    'fragment' => 'edit-c4m-vocab-geo',
+  ]);
+
+  $form['subscribe_c4m_follow_term']['edit_buttons'] = [
+    '#type' => 'markup',
```
mfernea: I think this can be set as type container with value.

project/profiles/capacity4more/modules/c4m/user/c4m_user_profile/c4m_user_profile.pages.inc
```diff
@@ -149,3 +149,426 @@ function c4m_user_profile_remove_account_form_submit($form, &$form_state) {
...
+function _c4m_user_profile_settings_form_groups(array &$form, array $digest_options, &$original_values, $account) {
+  $form['subscribe_c4m_follow_group'] = [
+    '#type' => 'fieldset',
+    '#title' => t('Your group/project notifications'),
+    '#collapsible' => FALSE,
+    '#collapsed' => FALSE,
+  ];
+
+  $form['subscribe_c4m_follow_group']['summary'] = [
+    '#type' => 'markup',
```
mfernea: Is type required if we have only #markup?

project/profiles/capacity4more/modules/c4m/user/c4m_user_profile/css/user_settings.css
```diff
@@ -0,0 +1,27 @@
+.user_settings_row {
```
mfernea: I would as this to the theme in order to:
- have access to variables and / or mixins
- use sass nesting

We also only have few separate css files within modules.

PR #1849, opened 2018-03-15
CFDRMD-250: Upgrade Travis distro + PHP 7.

soniCaH:


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: php
-dist: precise
```
mfernea: I think we should add dist: trusty to not have the same surprise as we did when they changed the default to trusty.

.travis.yml
```diff
@@ -15,10 +16,16 @@ env:
...
+  - 7.0
+  - 7.1
+  - 7.2
+  - nightly
 
-node_js:
-  - '4'
+matrix:
+  fast_finish: true
+  allow_failures:
```
mfernea: Should we test for 7.2 and nightly if we allow failures? Doesn't the job take too long then? Does the fast_finish setting makes the job successful even if it has to run on 7.2 and nightly?

ci/bin/install_drupal.sh
```diff
@@ -21,15 +18,14 @@ cp $TRAVIS_BUILD_DIR/ci/config/config.sh.travis config/config.sh
 ./bin/init -y
 
 # Install the drupal platform based on the config file.
-./bin/install -y --dummy-content --no-login --no-backup --env=ci
+./bin/install -y --dummy-content --no-login --no-backup --env=ci --verbose
+
+cd $TRAVIS_BUILD_DIR/web
```
soniCaH: Can be removed as we use aliases for drush.

PR #1851, opened 2018-03-16
CFDRMD-250: Travis build on Xenial (16.04)

soniCaH:


soniCaH:
Xenial is only supported end of April 2018.

PR #1895, opened 2018-06-21
Pinpoint the solr version drush needs to download.

soniCaH:
Default version was switched to Drupal 8 version, since we are not in a Drupal environment when we download the module. Therefore the config folder in the installation instructions was changed and caused errors.


soniCaH:
Replaces #1894

